PR #3, opened 2016-10-15
Added PSD for packed matrices

mfalt:
This enables us to use the packed format for symmetric matrices when working with the `IndPDS` using the BLAS function `dspev!`. Unfortunately there is currently no accepted type for packed matrices in julia Base so I suggest using this vector format until it is added to Base.
I am not sure however how this PR affect the documentation.
This PR only includes support for `Float64`, It is however possible to add single precision implementation using `sspev!`.


lostella:
I don't understand: it looks like there are some issues when importing `Base.LinAlg.BLAS.@blasfunc` in Julia 0.4, [see this build](https://travis-ci.org/kul-forbes/Prox.jl/builds/168284470). But previous builds did not show this, [see this other one](https://travis-ci.org/kul-forbes/Prox.jl/builds/168279765).

mfalt:
I tried another PR but it doesn't work either. It seems that julia have had both
`Base.blasfunc`, `Base.@blasfunc`, `Base.BLAS.@blasfunc` and now `Base.LinAlg.BLAS.@blasfunc` and that not even Compat is handling it correctly. Its probably not a priority for them since `blasfunc` is not an exported command.

PR #5, opened 2016-10-21
Test typestable

mfalt:
Should we consider having a type stable output like this?


lostella:
Sure, I guess this is performance-related?

mfalt:
Yes, before, every time `dspev!` was called, julia couldn't figure out the type of the output on compilation, because it didn't only depend on the type of the inputs, but also on the value of jobz

``` julia
jobz == 'V' ? (W, Z) : W
```

so any function that calls `dspev!` will have to check the type of the output. The function `dspevV!` is "type stable", i.e. the type of output is given from the type of input, therefore julia can compile efficient code for functions that call `dspevV!`.

PR #11, opened 2017-01-27
added Huber loss

lostella:


lostella:
Looks like tests fail on 0.6 because of malfunctioning of `svds` there. Merging.

PR #12, opened 2017-02-04
finished Precomposition, corrected HingeLoss, getting ready for 0.6

lostella:


src/functions/indBallRank.jl
```diff
@@ -41,7 +41,11 @@ function prox!{T <: Real}(f::IndBallRank, x::AbstractArray{T,2}, y::AbstractArra
     for j = 1:size(x,2)
       y[i,j] = 0.0
       for k = 1:f.r
-        y[i,j] += svdobj[:U][i,k]*svdobj[:S][k]*svdobj[:Vt][j,k]
+          if VERSION > v"0.5"
```
tkelman, line 44: you should be more specific about version cutoffs like this

lostella: Ok, thanks! I guess this change is necessary since [this](https://github.com/JuliaLang/julia/pull/19974), how should make the version cut more specific?

tkelman: check out the merge commit https://github.com/JuliaLang/julia/commit/dd6ceffd4efdc8254387d5061f0ee188fcb44dc9 and run the `contrib/commit-name.sh` script

lostella: Also: I noticed that many (if not all) PRs to METADATA.jl come with a nice comment by the author of the PR pointing to diffs, Travis builds and so on. I guess that is generated automatically when one publishes a package release directly from the Julia REPL, right?

tkelman: see https://github.com/attobot/attobot#usage

Issue #14, opened 2017-02-13
IndicatorConvexCone


mfalt:
You seem to have defined `IndicatorConvexCone <: IndicatorConvex` but most cones are defined as
`immutable IndPSD <: IndicatorConvex end`
`immutable IndFree <: IndicatorConvex end`
Is there a reason for this? I would be happy add a PR if not.
It would also be nice for the package I'm writing (and for slight efficiency) if we had `IndNonnegative` (and `IndNonpositive`) as `IndicatorConvexCone` instead of
`IndNonnegative() = IndBox(0.0, +Inf)`.
That way I would be able to restrict my algorithm to work with any `IndicatorConvexCone` in this package.

We might also consider adding a `IndDualCone{T<:IndicatorConvexCone}`. I already have an implementation for this in my code, and it should be easy to include here. It looks something like this:
```julia
type DualCone{T<:IndicatorConvex} <: IndicatorConvex
    C::T
end

dual{T<:IndicatorConvex}(C::T) = DualCone{T}(C)
dual(C::DualCone) = C.C

function prox!(cone::DualCone, x, y)
    prox!(cone.C, -x, y)
    for i = 1:length(x)
        y[i] = x[i] + y[i]
    end
end
```

and optionally we could add something like this
```julia
#Wrapper for dual prox to avoid double duals
function proxDual!(C::IndicatorConvex, x, y)
    prox!(C, -x, y)
    @simd for i = 1:length(x)
        y[i] = x[i] + y[i]
    end
end
proxDual!(C::DualCone, x, y) = prox!(C.C, x, y)
```

I'm happy to create a PR for this too if you are interested.

lostella:
Regaring your first point: no, there's no reason for that. I guess that's just a mistake and `IndPSD` and `IndFree` should in fact be of type `IndicatorConvexCone`.

The same applies to `IndNonpositive` and `IndNonnegative`. I just made them as they are so as to have a shortcut for their implementation. It makes perfect sense (also for performance) to specialize their implementation. One could keep track of their correctness by comparing them with their `IndBox`-based implementation in tests.

As for your last point (dual cone), check out [conjugate.jl](https://github.com/kul-forbes/ProximalOperators.jl/blob/master/src/calculus/conjugate.jl), which already implements Moreau identity for computing the `prox!` of conjugate functions.

lostella:
[...]

PR #17, opened 2017-02-14
Cones

mfalt:
Consistently consider cones as type `IndicatorConvexCone` according to #14. Added property to FUNCTIONS.md.


lostella:
So I guess this solves the first half of #14, the other half being restructuring the conjugate operation

PR #18, opened 2017-03-07
Fixed RotatedSOC and added tests

mfalt:
Fixed bug in `RotatedSOC` and added tests, for example
(p,q,x) = [1., 1., sqrt(2)] should be on the boundary of |x|^2 <= 2*p*q which it wasn't before.


mfalt:
The first build failed because I stopped travis (since I pushed a new version). The second failed because coverage decreased on a file I didn't change. Maybe because of random numbers in the tests? We should probably consider not using random numbers in the tests? Or at least set the seed.

PR #20, opened 2017-03-25
indBallL1: simpler code and no type instability

mfalt:
Avoids type instability and allows for simpler code.
`p` was previously of type `Array{Any,1}` and the compiler could not know (for a matrix input) if `p` would become a vector or a matrix.


lostella:
I closed it by mistake, sorry :-) I'm merging it asap, thanks!

Issue #15, opened 2017-02-13
Completing Precompose


lostella:
The `Precompose` calculus rule needs to be completed. Given `f` and a linear mapping `A`, we want to be able to compute the `prox!` of `x -> g(x) = f(Ax)`.

[As of now](https://github.com/kul-forbes/ProximalOperators.jl/blob/6a4c8535187d84f73be1df4aaf73c58182eb1392/src/calculus/precompose.jl) we are able to handle the case where `f` is precomposed with a diagonal mapping. In this case `A` is represented by `a`:

1. `a` is a `Real`: then the mapping is a multiple of the identity, in which case the rule can be applied regardless of the structure of `f`;
2. `a` is an `Array`: then `a` represents the mapping's diagonal, and the rule can be applied provided that `f` is separable (actually, if all elements of `a` are equal then this is the same of case 1., but we don't check that now); separability of `f` is detected by the presence of `prox!(f, x::Array, y::Array, gamma::Array)`, i.e., the possibility of specifying a different stepsize for each coordinate. See for example how the [hinge loss](https://github.com/kul-forbes/ProximalOperators.jl/blob/6a4c8535187d84f73be1df4aaf73c58182eb1392/src/functions/hingeLoss.jl) is defined in terms of the [sum-of-positive](https://github.com/kul-forbes/ProximalOperators.jl/blob/6a4c8535187d84f73be1df4aaf73c58182eb1392/src/functions/sumPositive.jl).

This structure presents a few drawbacks. I can think of:

- [x] Having a call with `gamma::Array` as fourth argument may create ambiguities: what if one wants to call the three-arguments `prox!`? Then `prox!(f, x::Array, gamma::Array)` clashes with `prox!(f, x::Array, y::Array, gamma::Real=1.0)`. **Remedy:** put the stepsize `gamma` as third argument, and the target (output) `y::Array` as fourth. This is backward incompatible but better now than never, and the more I think of it the more I'm convinced it's the right signature for `prox!`.
[...]

PR #22, opened 2017-04-03
New arguments order + separable functions

lostella:
This PR updates the interface in a substantial way, and partially addresses #15. It's a bit big, sorry about that.

1. The three-arguments `prox!(f, x, gamma)` is removed.
2. The arguments order of `prox!` changes from `prox!(f, x, y, gamma)` to `prox!(y, f, x, gamma)`. This aligns the syntax of `prox!` with that of other argument-modifying functions (such as `A_mul_B!`), putting the output-argument as first.
3. For separable functions (such as `NormL1`, `SqrNormL2`, `HingeLoss` and others) now last argument `gamma` is allowed to take vector values (of type `AbstractArray`) with dimension compatible with the point `x`.
4. A new predicate `is_separable` returns `true` if the function is separable, `false` otherwise. This is useful for randomized tests for example.
5. `Precompose` is now `PrecomposeDiagonal`, to differentiate it from the new calculus rule `PrecomposeGramDiagonal` which is added, but still incomplete and not yet exported.
6. Fixed types in `Conjugate`, `HuberLoss`, `LeastSquares`, `LogBarrier`, `SqrDistL2 `, `SqrNormL2`, `SumPositive`.

The critical items are 2 and 5.
* Item 2 requires updating packages dependant on ProximalOperators.
* Item 5 contains `PrecomposeGramDiagonal` which is yet to be completed: this should implement the composition of a function with a mapping `L` such that `LL'` is diagonal (with either uniform or non-uniform diagonal). The way I see it, the new function should be created given `L`, `L'` (the adjoint operator) and the diagonal of `LL'` (as a scalar if it's uniform, as a vector if it's non-uniform). Any ideas on what's the best way to provide `L` and `L'`? A `Function` may be fine, maybe an in-place function taking as first input an `Array` where to write outputs?

@mfalt and @nantonel let me know what you think about this, I'm fully open to reconsider some of these points. The important thing I guess is to have `gamma::Array` when possible (mainly concerning items 3 and 4), so as to have `PrecomposeDiagonal` and `PrecomposeGramDiagonal` with non-uniform diagonals (item 5).


mfalt:
I have no complaints, the changes look reasonable (gave the code a quick look), and I think 2 is a very good change.

nantonel:
Agree with the changes! 

I think the in-place function could be tricky to handle (but maybe needed for efficiency?) 
for example if we would have an in-place function `L!(y,x)`
instead of 
```
function (g::PrecomposeGramDiagonal){T <: RealOrComplex}(x::AbstractArray{T})
  return g.f(g.L(x) .+ g.b)
end
```
we would need to have a buffer `Array` and change this call into
```
function (g::PrecomposeGramDiagonal){T <: RealOrComplex}(b::AbstractArray{T}, x::AbstractArray{T})
  g.L!(b,x)  
  return g.f(b .+ g.b)
end
```
In order to keep the calls consistent, a buffer inside could be created inside `PrecomposeGramDiagonal`? 
```
immutable PrecomposeGramDiagonal{T ...} <: ProximableFunction
f::T
L::Function
Lb::AbstractArray
....
```
but probably there should be an easier way...

lostella:
Well you're right, but I guess that pre-allocating a buffer for storing the output of `L` requires that one knows what the range space of `L` is. Therefore only having `L::Function` and `Ladj::Function` may not do the job.

Issue #25, opened 2017-04-07
Using predicates instead of types


lostella:
Types in Julia are great. But types in Julia should not be overused. As pointed out by @mfalt, it is rather complicated to encode properties of `ProximableFunction` objects using Julia types: the simple fact that `f` is `ProximableConvex`, for example, is lost when constructing `Postcompose(f)`. What makes more sense, probably, are predicates of the form `is_convex(f)`, `is_cone(f)` and so on, that inspect object `f`. See #14 for a related discussion.

mfalt:
I completely agree, I think types should be used for what they are good for; dispatching functions and inferring types for efficient code.
I don't see that there would be any loss in efficiency by removing these abstract types,.

Issue #21, opened 2017-03-28
De-randomized tests


lostella:
As noted several times, code coverage bounces up and down due to randomized tests. One way of avoiding this is to hardcode all cases to be tested: time consuming, but allows to cover all borderline cases. A simpler way would be to initialize the RNG seed somewhere. Definitely not a major issue, but also serves as reminder that we should put order to tests at some point.

- [x] Set at least the RNG seed
- [ ] De-randomize tests

mfalt:
I think it may also be dependent on the order the tests on the different
platforms finishes. They seem to differ by the amount that is reported. I
haven't verified this though.
But we should make sure the tests are non random. As well as improving the
tests in student ways, such as looping over tests for different types, and
test that the results are of that right type.

PR #27, opened 2017-04-08
Removed types, introduced predicates

lostella:
This aims at solving #25. Now only `ProximableFunction` is there as type, and all other significant properties of functions are accessed at runtime via the following predicates: `is_separable`, `is_convex`, `is_set`, `is_cone`. The value of these predicates is defined for all basic types `T <: ProximableFunction` and is correctly propagated to composite objects constructed using calculus rules.


lostella:
I also added the very minor 950d606 which partially addresses #21 and will hopefully make coveralls give more green ticks for a while (I also disabled coveralls comments on PRs, so there should be less email flooding).

PR #29, opened 2017-04-16
More separable sum tests

lostella:
This PR basically just adds tests: it modifies a bit the way tests are performed (see test/runtests.jl lines 13, 29) and adds tests for the `SeparableSum` and `SlicedSeparableSum` types. It also adds deepArrays.jl that extends `copy!`, `similar`, `vecdot`, `vecnorm`, `maxabs` into respective `deepcopy!` etc., which are convenient for doing operations on block-split variables (i.e., `Array{Array}` types of variables).


lostella:
Need to add a few more commits to increase coverage before merging

PR #33, opened 2017-05-23
use version number instead of release in .travis.yml

tkelman:
release will change over time, but your REQUIRE file says this package supports julia 0.5 so it should continue to be tested

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,12 +2,6 @@ environment:
   matrix:
   - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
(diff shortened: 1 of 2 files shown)


lostella:
You're totally right @tkelman: I guess it also makes sense not to test against `nightly`, right?

tkelman:
depends whether you want to find out about things on prereleases for the sake of testing. could make it an allowed failure, which would run but not make your overall status red

appveyor.yml
```diff
@@ -5,6 +5,10 @@ environment:
   - JULIAVERSION: "julianightlies/bin/winnt/x86/julia-latest-win32.exe"
   - JULIAVERSION: "julianightlies/bin/winnt/x64/julia-latest-win64.exe"
 
+allow_failures:
```
tkelman: don't think this syntax worked on appveyor, I believe it needs a matrix like on travis

PR #35, opened 2017-05-26
using tuples in separable sums

lostella:
**Work in progress - do not merge**

This PR changes a bit how `SeparableSum` and `SlicedSeparableSum` work.

* `SeparableSum` now internally stores a `Tuple` of `ProximableFunction`; furthermore, the two points `x` and `y` to be passed to `prox!` are now also `Tuple`s. There are no other significant modifications to improve performance for the moment, but I guess from here we can improve things.

* `SlicedSeparableSum` now also uses `Tuple`s internally, and the `prox!` and function call are `@generated` as suggested by @mfalt so as to improve efficiency. Additionally, a constructor where just one function is specified for all slices has been added.


lostella:
Closing; will reopen. The modifications proposed here are now in the [tuples](https://github.com/kul-forbes/ProximalOperators.jl/tree/tuples) branch, which was branched from [julia0.6](https://github.com/kul-forbes/ProximalOperators.jl/tree/julia0.6).

PR #37, opened 2017-06-18
many new features

lostella:
Work in progress **DO NOT FORK/MERGE**

This PR contains several additions in many parts of the package. Although very rich, this PR is **almost fully backward compatible** (only one function, `FirmThreshold` has been removed from the library: all the rest are additions).

Need before merging:
- [x] Verify coverage and add missing tests.

Follows a summary of the major/minor changes.

**Major updates**

1. Added several predicates `is_xxx` to inspect properties of functions, see [ProximalOperators.jl](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/ProximalOperators.jl#L94) to see how their default value is set: each one of them is then specialized in the appropriate source file for each function type. Their name is supposed to be self-explanatory, but I will add comments/documentation on this.
1. Added `gradient!` and `gradient` for the functions `f` having `is_smooth(f) = true`. It makes sense to me to also offer the gradient operation when possible. Why not?
1. Added `Quadratic`, see [here](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/functions/quadratic.jl): this represents general quadratic functions given their Hessian and linear term. The `prox!` operation is implemented both directly as we do for `LeastSquares`, but also iteratively using a custom CG method: this can be found [here](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/utilities/cg.jl) and we could use it for `LeastSquares` as well, however it is very basic and we may think of relying on some better implementation.
1. Added `Linear`, see [here](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/functions/linear.jl): pretty self-explanatory.
1. Removed `FirmThreshold`: we lacked a rigorous definition of what the function was. We can always reintroduce it but with a proper name maybe, instead of the name of the proximal mapping (this it what "firm threshold" sounds like).
1. Added `MoreauEnvelope`, see [here](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/calculus/moreauEnvelope.jl). This only offers the `gradient!` operation. However, `MoreauEnvelope(f)` is smooth only when `is_convex(f) == true`.
1. Added `Precompose`, see [here](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/calculus/precompose.jl): this is a tentative to complete/generalize what was there already with `PrecomposeDiagonal`, addressing #15.
1. Added `Translate`, see [here](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/calculus/translate.jl).
1. Added **documentation** using Documenter.jl. This can be found in the docs/ folder: when this PR is merged I will add to the `gh-pages` branch what gets produced in the build/ folder, so for the documentation to appear online. I tried to organize what this package offers in a nice way, but of course it's only a first tentative and can surely be improved. For example: I didn't type at all the docstrings for readability purposes, and preferred instead to explain the nature of parameters in the text.
1. Related to the above point: updated many **docstrings**. It is very nice to put them in a way such that they automatically produce LaTeX formulas through Documenter.jl. However this may affect what they look like in the REPL.

**Minor updates**

1. Moved `DistL2` and `SqrDistL2` to the calculus rules folder.
1. Added credits to @mfalt! If you don't like this or need to customize it, please feel free to change it!
1. Added robust PCA example in the demos/ folder. It is also in the tests (and because of this tests take a bit longer to complete). I took this from Parikh and Boyd, "Proximal Algorithms".
1. I added the code for `Epicompose`, another calculus rule, [here](https://github.com/kul-forbes/ProximalOperators.jl/blob/gradient/src/calculus/epicompose.jl): this is not included yet, as it is still a work in progress. Will complete it.


mfalt:
Credits looks good, thanks!
I have my own implementation of `cg!` that I implemented this week, we can compare to that and see if we did something different.
I like that you have both direct and indirect solve for Quadratic, but maybe that should be an abstract type instead? The same could be done for `IndAffine`, I think I have working code for that. I will look through the code as soon as I have time, interesting changes!
The only thing i reacted to so far is `cg!(y, f.Q + I/gamma, x/gamma - f.q)`. This seems like it could be quite inefficient since it needs to allocate and compute a new matrix for every `prox!` call. I can see two options, either represent `f.Q + I/gamma` as a new type, and define `A_mul_B!` for it, or store `f.Q + I/gamma` in a specific `QuadraticIterative` type.

lostella:
Could be a different abstract type QuadraticIterative, yes, and I also like very much the idea of storing `Q + I/gamma` in a new type implementing `A_mul_B!`. Anyway this PR will stay there for a few days at least because I see coverage is pretty bad

PR #39, opened 2017-07-19
PSD scaling

mfalt:
Implements the scaling of PSD cone as described in Vandenberghe 2010: http://www.seas.ucla.edu/~vandenbe/publications/coneprog.pdf


mfalt:
This implementation was needed to integrate nicely with the [MathProgBase](http://mathprogbasejl.readthedocs.io/en/latest/conic.html) conic models. It shouldn't brake any existing code since IndPSD() defaults to the old implementation. If we pull this, it would be nice to tag a new version so that I can require it in my package.

lostella:
This affects only the calls to `prox!` with vector arguments `x` and `y`, right? Also, I'm wondering whether the scaling option should **always** be true for vector arguments (and get rid of the option)

lostella:
Related to my comment above: it would be nice to have a test where the two calls (with `Symmetric` and `Vector` arguments `x`, `y`) are compared on the same example. I guess only the **scaled** `Vector` version should agree with the `Symmetric` version.

lostella:
Sorry, I'll answer myself: of course the option should be there, it all depends on **how** a symmetric matrix is represented as a vector (whether with or without scaling). The option should stay there. @mfalt if you are happy with the PR feel free to merge it :-)

mfalt:
> This affects only the calls to prox! with vector arguments x and y, right?

Also the call `f(x::AbstractVector)`, otherwise, yes.
>Related to my comment above: it would be nice to have a test where the two calls (with Symmetric and Vector arguments x, y) are compared on the same example

The test I added is actually a copy of the standard PSD test, but with input and expected output scaled by `sqrt(2)` on the "off diagonal". Is this what you mean?

>  of course the option should be there, it all depends on how a symmetric matrix is represented as a vector 

Yes I agree, both ways are reasonable representations so its better to support both,

lostella:
What I mean is a test verifying (pseudo-julia coming):

`prox(f, vec(X)) == vec(prox(f, X))`
`prox(f, mat(x)) == mat(prox(f, x))`

where `vec` and `mat` are the transformations in Vanderberghe's notes (with and without scaling). But the test which is there is already OK: we can implement this more explicit "consistency" check later in time.

Feel free to merge, then I'll create a new tag immediately.

mfalt:
I see, but I think it is there indirectly, since previous test runs
```julia
prox(f, X) == Y
prox(f, x) == y
```
and the added tests, for the same, X,Y
```julia
prox(f, vec(X)) == vec(Y)
```

Issue #38, opened 2017-07-12
Project on operator graph and IndAffine row rank


mi-volodin:
Hello, @lostella !
I was looking for the reliable library for proximal operators evalutation on github and I liked this one because of the general architecture and implementation (👍🏻).

However, what I am missing was the projection to an operator graph case {(x,y) | Ax=y}. Since it also could be done via Affine projection - I decided to look through the code and discovered one interesting feature.
The implementation requires full rank matrices (so the QR method for Pseudo-inverse might be applied) which... might be my case.

So, I have the following questions:

1. was there any reason to do so (I mean deliberately prohibit) or it's just "not implemented yet" for the case m<n. In the least case I would try to participate and add this functionality (not sure it would be possible, but I'll try). Even SVD method might be suitable (since pinv could be stored inside the struct/immutable).

2. what do you think about adding the graph project case. It could be done through affine projection, but as far as I know it has it's own heuristics for efficient calculations.

mfalt:
I think these are very good points. I can only answer in regard to my own applications.

1. In the applications I'm working it's usually assumed that the matrix has full row rank. The other cases consist either of that no solution exist, in which case the projection is not defined, or that there are duplicate constraints, which could be removed in some pre processing step.

2. I have been working on an implementation of this, mainly an indirect method using some conjugate gradient. If you have any ideas on which algorithm to use I would be very happy to hear your input. My code is mostly in the [FirstOrderSolvers](https://github.com/mfalt/FirstOrderSolvers.jl) package, but I'm not sure if it contains a working version yet, or if it is correct. This, or another version, could probably be added/moved to this package.

mi-volodin:
Well, I'm interested in implementation of operator block-splitting algorithm, described [here](http://web.stanford.edu/~boyd/papers/block_splitting.html). I have already implemented it in MATLAB but ... there's a million "but"s about matlab. So now I'm doing the same thing in Julia.

1. In block splitting it is generally hard to think about row rank - the heuristic of how to pick up blocks has more importance. So, the case when `m<n` is normal. However, I asked here about it, because current implementation is way better for full row rank and SVD variant (see [wikipedia](https://en.wikipedia.org/wiki/Moore–Penrose_pseudoinverse), Construction) has several drawbacks.
But if it is not prohibited - I'd like to add this case implementation (if I could, pinv() of Julia does not accept sparse matrices, so I'll try to find a workaround).

[...]

PR #40, opened 2017-07-19
IndGraph.jl - projection on operator graph

mi-volodin:
Creating this as a starter for discussion which began in [Issue 38](https://github.com/kul-forbes/ProximalOperators.jl/issues/38)

@lostella currently this variant will work only for sparse matrices and raise an error for dense arrays. However, the method itself is applicable, though, not computationally efficient for dense cases.

As you kindly agreed to help me make this in correspondence with you product standards I'd like to ask first if you may suggest the desirable representation for two other cases: fat (m<n) and skinny (m>n) matrices.

The implementation is easy, but different. I was thinking about creating the three subtypes (sub-structures). Namely IndGraphSp, IndGraphFat, IndGraphSkin. This will help to dispatch prox!.
However, if I understand correctly IndGraph should then be an abstract type.

What do you think?

cc @mfalt 

```
abstract type IndGraph end

struct IndGraphSp <: IndGraph
     ....
end

struct IndGraphFat <: IndGraph
   ....
end

function IndGraph(A::AbstractArray{T,2} )
   if issparse(A)
       return IndGraphSp(...)
   else
       return IndFraphFat(...)
   end
end
```


src/functions/IndGraph.jl
```diff
@@ -0,0 +1,83 @@
...
+  end
+  return +Inf
+end
+
+function prox!{T <: RealOrComplex}(
+    x::AbstractArray{T, 1},
+    y::AbstractArray{T, 1},
+    f::IndGraph,
+    c::AbstractArray{T, 1},
+    d::AbstractArray{T, 1})
```
mfalt: This breaks the current consistency we have on all `prox!` signatures. I understand why this approach is desirable, but maybe there is another option? A special type for `x,y` would be one option, 
```julia
prox!{T}(xy::Tuple{AbstractVector{T}, AbstractVector{T}}, f::IndGraph, cd::Tuple{AbstractVector{T}, AbstractVector{T}}, gamma::Real)
```
is another. I'm not sure which is best though.

src/functions/IndGraph.jl
```diff
@@ -0,0 +1,83 @@
...
+  res = [c + f.At * d; zeros(f.m)]
```
mfalt: Is `f.At*d` really faster than `f.A'd` to warrant the extra storage? We should also consider trying to make this line and the others allocate as little memory as possible. Storing a persistant `tmp` and `xy` in  `IndGraph` could achieve this if you use the low level functions like `At_mul_B` with `view`s.

mfalt:
How is this approach different than creating the `IndAffine([A -I],zeros(n))` type? Is LDLt faster on `[A I;A' -I]` than QR on `[A -I]'`?
Regarding the abstract type, I have no problems with the types being abstract. I think we are consistently trying to make the `prox!` and evaluations type stable, but the constructors for the functions are not as important.

Edit: Don't forget to write tests, they are quite useful.

src/functions/IndGraph.jl
```diff
@@ -0,0 +1,83 @@
...
+# if !isdefined(:RealOrComplex)
+  RealOrComplex = Union{Real, Complex}
+# end
+#
+struct IndGraph{T <: RealOrComplex} <: ProximableFunction
+  m::Int
+  n::Int
+  A::AbstractArray{T,2}
+  At::AbstractArray{T,2}
+  F::Base.SparseArrays.CHOLMOD.Factor{Float64}
```
mfalt: This should break if `T` is `Complex` right?

src/functions/IndGraph.jl
```diff
@@ -0,0 +1,83 @@
+# indicator of an affine set
+
+"""
+  IndGraph(A::Array{Real,2})
+
+Returns the function `g = ind{(x, y) : Ax = y}`.
```
mfalt: It seems like the implementation could easily also handle `A*x+b=y`, maybe its worth to add that?

mi-volodin:
Regarding tests - yes, sure, I'm kind of a newbie in the github development process, so I'm unfamiliar with codecov.io, but I was going to make the tests anyway.
And - still the codecov checks are very helpful here :)

Regarding the question. I don't know how to answer it yet. The drawback of affine projection is the need to repack the `x` and `y` every time and, anyway, current limitation on matrix form `(m<n)`. I followed the recommendations in the paper mentioned in the Issue and made a special case for sparse matrices. However, I'm curious about computational properties of this approach, I'll try to find it out.

For the dense calculations... I'm not sure yet, but it looks very similar. The point is that if you know that the operator has the structure `[A -I]` -- you might use an heuristic to simplify computations.

So, to put everythin in a nutshell
* Math is going to be checked and motivation will be added here
* I'll implement it as I proposed and...
* Write the tests

And then we will come back to this question again :)

mfalt:
I have some code that uses the structure of the problem to solve it in an iterative way. I think that this could eventually go in this package, maybe as a part of the code you are writing. It may contain some tips on how to work with the views and special structures too: https://github.com/mfalt/FirstOrderSolvers.jl/blob/master/src/utilities/affinepluslinear.jl
In this case, I have both `x` and  `y` in a single variable without any substantial cost, and they are easy to work with, for example:
```julia
A = randn(m,n)
# Here I define f=ind([A -I][x;y]=0), no linear term
f = AffinePlusLinear(A, zeros(m), zeros(n), 1) 

xy = randn(m+n)
cd = randn(m+n)
x = view(xy, 1:n); y = view(xy, (n+1):(m+n))
c = view(cd, 1:n); y = view(cd, (n+1):(m+n))
#From here on, x,y,c,d should work as usual
# they are basically pointers to the location in xy and cd

x .= randn(n) #Do your stuff as if you are working on x, y
y .= randn(n)

prox!(cd, f, xy) #Works with the values set in x,y
```

Edit: I have still not figured out if CG is legal to use on the problem `[I A'; A -I]`, if anyone knows and have any references, I would definitely appreciate it.
Edit2: Note that the `prox!` allocates `O(1)` data each iteration by not introducing any temporary vectors.

src/functions/indGraphFat.jl
```diff
@@ -0,0 +1,58 @@
+# This implements the variant when A is dense and m < n
+
+struct IndGraphFat{T <: RealOrComplex} <: IndGraph
+  m::Int
+  n::Int
+  A::AbstractArray{T,2}
+  AA::AbstractArray{T,2}
+  F::Base.LinAlg.Cholesky #LDL factorization
```
mfalt: `AbstractArray{T,2}` and `Base.LinAlg.Cholesky` are not a concrete types, thus julia will not be able to infer from the type `IndGraphFat{T}` which methods to call when using these variables. Since we already know that we are working with dense matrices it should be safe to let the types be
```julia
A::Array{T,2}
AA::Array{T,2}
F::LinAlg.Cholesky{T,Array{T,2}}
```
unless we want the user to be able to supply their own versions of dense matrices, in which case we let them be of type `T1,T2,T2`.

src/functions/indGraphFat.jl
```diff
@@ -0,0 +1,58 @@
...
+# is_set(f::IndGraph) = true
+# is_cone(f::IndGraph) = false
+
+function prox!{T <: RealOrComplex}(
+    x::AbstractArray{T, 1},
+    y::AbstractArray{T, 1},
+    f::IndGraphFat,
+    c::AbstractArray{T, 1},
+    d::AbstractArray{T, 1})
+  y[:] = f.F \ (f.A * c + f.AA * d)
```
mfalt: Here it is probably good to have a `tmp` vector allocated in IndGraphFat if we want to avoid allocating memory in prox! , then we could do something like
```julia
A_mul_B!(tmp, f.A, c) # Maybe tmp .= f.A * c works equally well? Try it! 
A_mul_B!(y, f.AA, d)   # Maybe tmp .= f.AA * d
tmp += y
A_ldiv_B!(y, f.F, tmp)
```

src/functions/indGraphFat.jl
```diff
@@ -0,0 +1,58 @@
...
+  At_mul_B!(x, f.A, d - y)
+  x[:] += c
```
mfalt: I dont think the colon adds anything here, `+=` already signals "update x`

src/functions/indGraphFat.jl
```diff
@@ -0,0 +1,58 @@
...
+  return 0.0
+end
+
+function (f::IndGraphFat){T <: RealOrComplex}(x::AbstractArray{T,1}, y::AbstractArray{T, 1})
+  # the tolerance in the following line should be customizable
+  if norm(f.A * x - y, Inf) <= 1e-10
```
mfalt: We could use the same tmp vector here to avoid allocations.

src/functions/indGraphFat.jl
```diff
@@ -0,0 +1,58 @@
...
+# fun_expr(f::IndGraph) = "x,y ↦ 0 if Ax = y, +∞ otherwise"
+# fun_params(f::IndGraph) =
+#   string( "A = ", typeof(f.A), " of size ", size(f.A))
+
+function prox_naive{T <: RealOrComplex}(
+    f::IndGraphFat,
+    c::AbstractArray{T, 1},
+    d::AbstractArray{T, 1})
+
+  y = f.F \ (f.A * c + f.AA * d)
```
mfalt: Its all right to have allocations here, this can be as slow as it wants, as long as its readable.

src/functions/indGraphSkinny.jl
```diff
@@ -0,0 +1,54 @@
+# This implements the variant when A is dense and m > n
+
+struct IndGraphSkinny{T <: RealOrComplex} <: IndGraph
+  m::Int
+  n::Int
+  A::AbstractArray{T,2}
+  AA::AbstractArray{T,2}
+  F::Base.LinAlg.Cholesky #LDL factorization
```
mfalt: Same thing with Abstract types here

src/functions/indGraphSkinny.jl
```diff
@@ -0,0 +1,54 @@
...
+end
+
+function IndGraphSkinny(A::AbstractArray{T,2}) where {T <: RealOrComplex}
+  m, n = size(A)
+  AA = A' * A
+
+  F = LinAlg.cholfact(AA + speye(n))
```
mfalt: Sparse eye seems weird, I assume julia will convert it to dense (since AA is dense) and then add them. We might want to avoid allocating so much memory here, but it might not be important. The constructors are not assumed to be called to often.

src/functions/indGraphSkinny.jl
```diff
@@ -0,0 +1,54 @@
...
+end
+
+function prox!{T <: RealOrComplex}(
+    x::AbstractArray{T, 1},
+    y::AbstractArray{T, 1},
+    f::IndGraphSkinny,
+    c::AbstractArray{T, 1},
+    d::AbstractArray{T, 1})
+
+  x[:] = f.F \ (c + f.A' * d)
```
mfalt: Same comments as for fat above

src/functions/indGraphSparse.jl
```diff
@@ -0,0 +1,56 @@
+struct IndGraphSparse{T <: RealOrComplex} <: IndGraph
+  m::Int
+  n::Int
+  A::AbstractArray{T,2}
```
mfalt: Abstract type again

src/functions/indGraphSparse.jl
```diff
@@ -0,0 +1,56 @@
...
+function prox!{T <: RealOrComplex}(
+    x::AbstractArray{T, 1},
+    y::AbstractArray{T, 1},
+    f::IndGraphSparse,
+    c::AbstractArray{T, 1},
+    d::AbstractArray{T, 1})
+  res = [c + f.A' * d; zeros(f.m)]
+  xy = f.F \ res
+  x[:] = xy[1:f.n]
+  y[:] = xy[end - f.m + 1:end]
```
mfalt: for these lines there are a few options to allocate less (almost no) memory, for example something like letting `res, xy` be allocated vectors in IndGraphSparse, and then
```julia
res1 =  view(res,1:n) #This is basically a pointer to only the first part of res, the other part should always remain zero
At_mul_B!(res1, f.A, d) # This sets the first part in res
res1 += c

A_ldiv_B!(xy, f.F, res1)

x .= view(xy,1:f.n) 
y .= view(xy, (f.n+1):(f.n+f.m))
```
I am not sure if the two last lines allocate memory as you wrote them, sometimes the syntax `xy[range]` on the right side does that. 
There are other alternatives that might be equally good as the two last lines i wrote, maybe more or less efficient or readable, like
```julia
copy!(x, 1, xy, 1, n)     #Copy to x starting at 1, from xy starting at 1, n elements
copy!(y, 1, xy, f.n+1, m) #Copy to y starting at 1, from xy starting at f.n+1, m elements
```
or using the `@views` macro which means the lines you wrote can be written
```julia
@views x .= xy[1:f.n]
@views y .= xy[end - f.m + 1:end]
```
which will transform into the view example I gave. This will only allocate a "view", like a pointer which is a constant and small allocation, the copy! should be the most efficient however.

test/test_graph.jl
```diff
@@ -0,0 +1,70 @@
...
+
+prox!(x, y, fiagdf, c, d)
+@test fiagdf(x, y) < 1e-10
+@test norm(x_a - x, Inf) < 1e-10
+@test norm(y_a - y, Inf) < 1e-10
+
+prox!(y, x, fiagds, v, w)
+@test fiagds(y, x) == 0.0
+
+@time for i = 1:10000
```
mfalt: These kind of tests are good to do on your own computer to make sure that you write efficient code, but we don't really need them in the package since we don't monitor the time in the tests, only the results. We have been thinking about writing some benchmark tests, but I don't think we have anything yet. Take a look at the other tests in the package to see how we are doing to test our other methods. I think all the code in this file could be moved into those automated tests.

mfalt (commented):
Good work, I like the setup in general, but there are a couple of things that we should fix. I think I made a couple of comments when you first posted this PR, so they might be outdated.

EDIT:
I did some extra test, it seems like `tmp .+= y` is actually slightly better than `tmp += y`, so you might want to do that instead when considering my comments.

src/functions/IndGraph.jl
mfalt: I am not sure if this is true anymore

mfalt:
The tests failed because you wrote `include("functions/indGraph.jl")` but the file is called "functions/IndGraph.jl". It is easy to test the whole package, basically the same as the automatic tests, on your own computer by running `Pkg.test("ProximalOperators")`.

mi-volodin:
@mfalt That is strange. I did the Pkg.test() before commiting and all went ok. And file is actually `indGraph.jl`. Maybe it was a commit error, I'll fix it.

mfalt:
Are you running Windows? It seems like the windows code passed (windows is not case sensitive if I remember correctly).
GitHub says that the files are IndGraph.jl, indGraphFat.jl, indGraphSkinny.jl

src/functions/IndGraph.jl
mi-volodin: Hmmm... I doubt that. Unless the KKT system remains quasidefinite. If I have some time I will check the math.

src/functions/IndGraph.jl
mi-volodin: Yes, this is a mistake. Thanks!

src/functions/IndGraph.jl
mi-volodin: Well, I'm a newbie here, so I do not know if tuple allow to change it objects, if yes - no problem. I can always make
```julia
prox!(x,y,F, c, d) = prox!((x,y), F, (c,d))
```

But I do not really want to add special types if it could be handled without it.

src/functions/IndGraph.jl
mi-volodin: Agree. In fact there are lots of optimization steps to be done. I just wanted to add a baseline with which everything might be compared :)

src/functions/indGraphFat.jl
mi-volodin: Yes, thanks, and the same with Sparse case.

src/functions/indGraphFat.jl
mi-volodin: Why don't use `y` as a container? Not faster but with no `tmp` variable
```julia
A_mul_B!(y, f.A, c) 
y[:] .+= f.AA * d # I think this is generally the same, as allocating tmp, but with no tmp variable
A_ldiv_B!(y, f.F, y) # or \= ?
```

src/functions/indGraphFat.jl
mi-volodin: Ok, I thought about that yesterday, but decided not to change :) Thanks!

mi-volodin:
Nope, OS X. That is strange. In my local repo everything is alright, but remote repo is `IndGraph.jl`. I will fix it somehow...

src/functions/indGraphFat.jl
mfalt: There is no efficient implementation for doing matrix-vector multiplication and adding the result to a vector, only writing it to a vector, so the line `y[:] .+= f.AA * d` will do something like
```julia
tmp = f.AA * d
y[:] .+= tmp
```
However, if we already allocate tmp and store it in `IndGraph`, julia won't have to allocate it every time `prox!` is called, it is only allocated when `IndGraph` is created and then reused. Wherever I wrote `tmp` in my code, it should be `f.tmp` instead.

mfalt: Try yourself with the code
```julia
struct IndGraphFat{T <: RealOrComplex} <: IndGraph
  m::Int
  n::Int
  A::Array{T,2}
  AA::Array{T,2}
  tmp::Array{T,1}
  F::Base.LinAlg.Cholesky #LDL factorization
end

function IndGraphFat(A::AbstractArray{T,2}) where {T <: RealOrComplex}
  m, n = size(A)
  AA = A * A'

  F = LinAlg.cholfact(eye(m) + AA)
  #normrows = vec(sqrt.(sum(abs2.(A), 2)))

  IndGraphFat(m, n, A, AA, Array{T,1}(m), F)
end

function prox!{T <: RealOrComplex}(
    x::AbstractArray{T, 1},
    y::AbstractArray{T, 1},
    f::IndGraphFat,
    c::AbstractArray{T, 1},
    d::AbstractArray{T, 1})
  A_mul_B!(f.tmp, f.A, c)
  A_mul_B!(y, f.AA, d)
  f.tmp .+= y
  A_ldiv_B!(y, f.F, f.tmp) # y .= f.F \ (f.A * c + f.AA * d)

  At_mul_B!(x, f.A, d - y) #This line can be more efficient, like using tmp to store (d-y)
  x .+= c
  return 0.0
end
```

src/functions/indGraphSparse.jl
mi-volodin: This might be a problem. Let us assume someone iterate with `indGraph` and `prox!` - someone like me ;)

So before I start I allocate `x` and `y` holders outside prox operator. If I understood your suggestions right - `view` variant lead to the situation when data physically stored inside `res` of `IndGraph` object. This is bad.

Regarding the `copy` variant - is there any difference between `x[:] .= xy[1:n]` and `copy!(x, 1, xy, 1, n)`? I assume yes, but can you point out the difference?

test/test_graph.jl
mi-volodin: Yep, agree, forgot to purge it.

src/functions/indGraphSparse.jl
mfalt: I did some tests on this line with n=500.
The `copy!` function only does a simple check for the dimensions and then calls an efficient C function that loops over the elements and adds them, allocating almost nothing. `@benchmark` from BenchmarkTools estimates :
79.601 ns and 0 bytes allocated. Memory scales as O(1)

The code  `x .= xy[1:n]`  or `x[:] = y[1:n]` allocates a temporary vector for the result of `xy[1:n]` and then copies  that to `x`, while doing bounds checks in both stages.
Takes 527.463 ns, allocates 4.09 KiB. Memory scales as O(n)

 `x[:] .= xy[1:n]` is similar but also creates a view of x first, seemingly slowing it down considerably.
Takes 2.791 μs and 4.47 KiB. Memory scales as O(n)

`@views  x .= xy[1:n]` or equivalent `x .= view(y,1:n)` creates a view of 1:n in y and then does efficient copy.
Takes 1.916 μs and 592 bytes. Memory scales as O(1)

`view(x,1:n) .= view(y,1:n)` and `view(x,:) .= view(y,1:n)`
Takes roughly 3.5 μs with 1 KiB memory, but scales as O(1) in memory.

mi-volodin: looks like `copy!` is the best variant :) agree?

mfalt:
Agree.

src/functions/indGraphSkinny.jl
mi-volodin: I have a question regarding this line and indicator function calculation. 
Since there is `c` dimension of `n < m` I require `Array(n)` here. But in indicator function I still need `Array(m)`. If `m` is just slightly bigger than `n` - I just use one `Array(m)` for indicator and a view on first `n` of it for `prox!`. 

However, in case `m >> n` (which should not occur, but everything is possible :) ) keeping `Array(m)` will allocate unnecessary memory, since indicator is used less frequent than `prox!`.

So, I have two variants (maybe there's a better way):
1. Use `Array(m)` and write both `prox!` and indicator in an efficient way
2. Use `Array(n)` and allow allocation on each indicator call.

@mfalt what do you think?

mi-volodin:
Tuples are not done yet. I just want to clarify if this approach is ok (see`IndGraphSparse.jl`).

Also, maybe this is something wrong with me, but 
`A_ldiv_B!(f.res, f.F, f.tmp)` is not working because there is no appropriate method... strange.

mfalt:
Looking at the source, it seems like the solve is done in-place in the input vector: https://github.com/JuliaLang/julia/blob/master/base/linalg/cholesky.jl
So I haven't tested, but this should work
```julia
A_mul_B!(f.tmp, f.A, c)
A_mul_B!(y, f.AA, d)
y .+= f.tmp
A_ldiv_B!(f.F, y)
```

src/functions/IndGraph.jl
mfalt: Yes, it's not possible to change what the elements in the tuple point to, but you are able to change the values in the arrays for example. The code you posted should work fine,  and a wrapper in the other direction should work fine too.

src/functions/indGraphSkinny.jl
mfalt: I think you are raising a valid point, but I think case 1 should work fine. Even if m>>n it is just a single allocation of size m, the matrix is already dense m*n. But it's not an obvious choice.

src/functions/indGraphFat.jl
```diff
@@ -3,40 +3,53 @@
...
 
 # is_convex(f::IndGraph) = true
 # is_set(f::IndGraph) = true
 # is_cone(f::IndGraph) = false
 
 function prox!{T <: RealOrComplex}(
-    x::AbstractArray{T, 1},
-    y::AbstractArray{T, 1},
+    x::Array{T, 1},
+    y::Array{T, 1},
```
mfalt: There is no reason to restrict the input arguments to be the array type, this would inhibit views as input for example, the same goes for the other functions.

mfalt: AbstractArray{T,1} is fine

src/functions/indGraphSparse.jl
```diff
@@ -1,40 +1,58 @@
-struct IndGraphSparse{T <: RealOrComplex} <: IndGraph
+struct IndGraphSparse{T <: RealOrComplex, Ti <: Int} <: IndGraph
```
mfalt: I think Int is an alias for Int64 or Int32 depending on the system, either restrict it to the same as in the definition of SparseMatrixCSC or don't restrict it at all.

src/functions/indGraphSparse.jl
```diff
@@ -1,40 +1,58 @@
...
 
 # is_convex(f::IndGraph) = true
 # is_set(f::IndGraph) = true
 # is_cone(f::IndGraph) = false
 
-function prox!{T <: RealOrComplex}(
-    x::AbstractArray{T, 1},
-    y::AbstractArray{T, 1},
+function prox!(
+    xy::Tuple{AbstractVector{T}, AbstractVector{T}},
```
mfalt: Does this work? If not, look into how subtyping works, i think the old approach you had works fine if we just add a wrapper.
I don't have access to a computer now but I would guess it should be something like `prox!{T <: RealOrComplex, T1<:AbstractArray{T}, T2<:AbstractArray{T}}(xy::Tuple {T1,T1}, f::IndGraphSparse, cd::Tuple {T2,T2})`
Or using the where syntax and/or letting x,y,c,d be four different types. Sorry for being brief, I am writing on my phone.

src/functions/indGraphFat.jl
mi-volodin: Yep :) I was mistakingly turn them to `Array` after your comment about concrete types for matrices. Have not roll it back in this file, certainly should have...

src/functions/indGraphSparse.jl
mi-volodin: It's work. By the way - is there any difference if I will keep my original functions without tuples (they are a bit more readable) and make a tuple wrapper OR vice versa (how it is implemented in last commit in Sparse variant)?

mfalt: I think it looks nicer to have it as `prox!(x, y, f, c, d)`,`any other call can be converted to this form.

mi-volodin:
Regarding the LDLt solution for sparses.
The source [https://github.com/JuliaLang/julia/blob/master/base/sparse/cholmod.jl](https://github.com/JuliaLang/julia/blob/master/base/sparse/cholmod.jl) shows that there is no 'inplace' method. All packed in `ccall` to SuiteSparse. 
So I try to soften the damage by the `.=` assignment

mi-volodin:
One more thing.
There are some predicate functions like `is_convex` or `is_cone`.
I'm not sure I understand the meaning of them. For example `is_convex` - is it regarding indicator function or proximal mapping of it? 

Will appreciate any hints!

mfalt:
Yes, you are right, I must have been looking at something else. I am looking into extending the CHOLMOD functionality in julia to wrap some extra C functions, but until then you can use the allocating version. If you are interested, I have an alternative package (https://github.com/mfalt/SparseLDLT.jl) for doing LDLt factorizations which supports `A_ldiv_B!`, feel free to try it, but it is not registered yet so you have to do `Pkg.clone("https://github.com/mfalt/SparseLDLT.jl.git")`. It may be a bit faster. Once you have done `using SparseLDLT` it will overload the `ldltfact` and add `A_ldiv_B!`. We will probably not add it to this package unless the improvement is substantial.
 
Regading `is_convex` and `is_cone`: In this case it doesn't really matter since projection onto an affine space is an affine operation and thus convex, but in general it a question about the function, not the prox, i.e., in this case `IndGraph`. The indicator function for an affine set is convex (since it is the indicator of an convex set), and since it is the indicator of a linear graph, (0 is in the set), it is also a cone.

mi-volodin:
@mfalt  How about adding such ldlt to the core? :)

mi-volodin:
Checks passed.
@mfalt Any additional adjustments or tests?

mfalt:
> How about adding such ldlt to the core? :)

I will if I'm able to create something better and they want it.

> Checks passed.
Any additional adjustments or tests?


I'll look through the code one more time, but it seems almost done, however, I wrote this comment a while ago:
> Take a look at the other tests in the package to see how we are doing to test our other methods. I think all the code in this file could be moved into those automated tests.

In particular, add one example of each type to test_calls.jl and one of each to test_results.jl. For test results you can either create a simple example where the solution can be verified by hand, or a tricker one where you get the correct answer from some other method like IndAffine. An example with complex values would be really nice to verify.

This would probably catch/verify the following things among other:
* You are not accepting the gamma parameter. (It won't affect the algorithms, but we always want to be able to call `prox!(y, F, x, gamma)`.
* If any of the outputs have the wrong type
* That the `f(prox(f,x))==0.0`

I'm guessing that if you add these tests you can probably remove the whole test file you wrote.

mi-volodin:
Agree. Thanks!

What about `gamma`? I know it won't affect the algorithms so I didn't add it. Would the additional signature be fine?

e.g.
```julia
function prox!(x,y,f,c,d)
   ...
end

prox!(x,y,f,c,d; gamma) = prox!(x,y,f,c,d)
prox!(xy,f,cd; gamma) = prox!((xy[1], xy[2]), f, (cd[1], cd[2]); gamma)
```
not sure about the syntax with `;` in call, but it should cover all cases.

EDIT: Or I can add `; gamma` to the definition of `prox!`.

mfalt:
The semicolon implies that it is a keyword argument, i.e. that it would be called as `prox!(x, f, y, gamma=1.0)`, that is not the syntax we are using in the package. In the rest of the code we have just defined
```julia
function prox!(y::AbstractArray{T,1}, f::SomeType, x::AbstractArray{T,1}, gamma::R=one(R)) where {R<:Real, T<:RealOrComplex{R}}
...
end
```
i.e. requiring that gamma is real, and of the same "complexity" as `x` and `y`, while defaulting to 1. In this case we could simply add the "gamma=1.0" to all the prox! definitions since we won't use it, i.e. define
```julia
function prox!(x,y,f,c,d,gamma=1.0)
   ...
end
```
which means that `gamma` is optional, and will be set to 1.0 if not supplied.

By the way, I think an even nicer wrapper than the tuple would be:
```julia
function prox!(xy::AbstractArray{T,1},f::IndGraph,cd::AbstractArray{T,1}, gamma=1.0) where {T<:RealOrComplex}
 @assert length(xy) == f.m+f.n
 @assert length(cd) == f.m+f.n
 x = view(xy, 1:f.n)
 y = view(xy, (1+f.n):(f.n+f.m))
 c = view(cd, 1:f.n)
 d = view(cd, (1+f.n):(f.n+f.m))
 prox!(x, y, f, c, d, gamma)
 return 0.0
end
```
Then both IndAffine and IndGraph could be used in the same way. Also note that since all subtypes of IndGraph have the fields f.m and f.n, this only has to be defined for the abstract type IndGraph, not for the individual concrete types.

mi-volodin:
>Also note that since all subtypes of IndGraph have the fields f.m and f.n, this only has to be defined for the abstract type IndGraph, not for the individual concrete types.

How? I only know subtyping from abstract types. I couldn't find anything about subtyping from composite type (`struct`). And defining anything in a type makes it non-abstract (in Julia). Maybe I'm not right, but this won't work
```julia

struct AA
   m::Int
end

struct BB <: AA #error - can't subtype composite type
   n::Int
end

abstract type CC # error - invalid abstract type definition
   m::Int
end
```

>By the way, I think an even nicer wrapper than the tuple would be

Okay, I add it. And... tuples won't be required?

>In this case we could simply add the "gamma=1.0"

Agree - will do.

mfalt:
Sorry for being unclear, what I meant is that the `prox!` as I wrote it only needs to be defined in the IndGraph file, assuming that the concrete types have fields `m` and `n`, which they do.

> Okay, I add it. And... tuples won't be required?

Yeah, skip the tuples, I don´'t see them being used anyway.

mi-volodin:
@mfalt 
As we discussed, todo:

- Add hard-coded example to test_results for each type of IndGraph. (Or, alternatively - generate "random" with the fixed seed).
- Remove commented unnecessary blocks.

Something else?

mi-volodin:
@mfalt @lostella if the checks pass, I'm pleading for a review.
Maybe, if it's ok, you might even consider the possibility of merging 😅 

EDIT: Found an issue about randomized tests... ehhm... I did `srand(time_ns())` in my `test_graph`...

test/runtests.jl
```diff
@@ -103,4 +103,8 @@ end
   include("test_demos.jl")
 end
 
+@testset "IndGraph vs IndAffine" begin
+  include("test_graph.jl")
```
lostella: You can probably include this in the `@testset "Functions" begin [...] end` block, instead of having its own `@testset` block.

src/functions/indGraph.jl
```diff
@@ -0,0 +1,79 @@
+export IndGraph
+
+"""
+**Indicator of an operator graph**
```
lostella, line 4: This docstring shall be included in the documentation. To do so, it is sufficient to add
```
IndGraph
```
into a `@docs` block in the appropriate markdown file. I think in this case the right place is in `docs/src/functions.md` in the `## Indicators of sets` section (note that docstrings are included in alphabetical order):
```
[...]
IndBox
IndGraph
IndHalfspace  
[...]
```

test/test_graph.jl
```diff
@@ -0,0 +1,56 @@
...
+x_a = @view xy[1:n]
+y_a = @view xy[n+1:end]
+
+## IndAffine QR
+prox!(xy, fiaf, cd)
+@test fiaf(xy) == 0.0
+
+## SPARSE
+# test call
+prox!(x, y, fiag, c, d)
```
lostella: It's probably always better to *also* test the 3-arguments call `prox!(xy, f, cd)` and the 1-argument call `f(xy)` (this applies also to the fat case, lines 52-53), since that is the "canonical" interface which is common to all types in the package and documented. I know the other test scripts `test_calls.jl` and `test_results.jl` do this, but just for completeness I would include this also here.

lostella (changes requested):
Just really minor changes before merging, as far as I'm concerned

test/runtests.jl
mi-volodin: Sure, I'll do

src/functions/indGraph.jl
mi-volodin: Okay, sure, I was quite ignorant about how doc works, so I just copied the description from your `IndAffine` PR and thought that it might be enough :)

test/test_graph.jl
mi-volodin: I didn't get it.
I mean, I've included all cases in `test_calc` and one hardcoded case in `test_results`. And since it is done by `prox_test` and other framework functions it uses exactly `prox!(xy, f, cd, gamma)` signature.

I had manual calls in previous commits, but they are less comprehensive than `prox_test`, so I cleaned it out.

lostella:
I'm really a fan of the "tuple" way of thinking "points" in the case of this function. That is, for a linear operator `A`, the graph is `G = {(x, y) : Ax = y}`, where `x` and `y` are vectors (that is, `Vector`s). I would have tried going for the "double interface" accepting either the `Tuple` object `(x,y)` as input (to `prox!` for example) or the "stacked" `Vector` object `[x; y]`, rather than the 5-arguments `prox!(x, y, f, c, d)`. But maybe we can add this in the future.

mi-volodin:
@lostella 
We already tried tuple case with @mfalt. 
There is not problem to get them back (I had them, it was deleted in later commits). I just would like to see a consensus with Mattias [see post](https://github.com/kul-forbes/ProximalOperators.jl/pull/40#issuecomment-320094709)

But I was personally interested about having `prox!(x, y, f, c, d)` because in my case `x` and `y` are subarrays or elements of two different objects.  So `[x;y]` variant is totally not my case, it would be very inefficient. And calculations with separate arguments are a bit more readable than with tuples.

But I see no problem in having all kind of methods for this case (and add the tuple variant).
Since `f` has a `ProximableFunction` supertype I don't expect any ambiguity in dispatching the other signatures...

test/test_graph.jl
lostella: Yes, what I'm suggesting is for example to use the `prox_test` (with `xy` and `cd` as inputs) and `call_test` also here: this will do all the checks, compare with the `_naive` implementations. All this will be done by at some point calling the 5-arguments `prox!(x, y, f, c, d)`, which is what is tested here now.

The reason for this is: I would like to clean up at some point the way tests are performed. The `test_calls` and `test_results` scripts are getting a bit messy and difficult to handle. I would rather, in the future, have one test script per function constructor, which uses `prox_test` and `call_test` for all routine checks, and then checks anything one may want to check: optimality condition of the computed proximal point, hand-curated cases with known expected outputs, equivalence with other functions (such as in this case, equivalence between `IndGraph` and `IndAffine` when appropriately used), and so on.

Therefore, say tomorrow we comment out `test_calls.jl` and `test_results.jl`, this remaining test script for `IndGraph` will not cover the call `prox!(xy, f, cd)`, but only the `prox!(x, y, f, c, d)`.

lostella:
@mi-volodin Yes in fact, it is really something that we can add later. Of course having the `prox!(x, y, f, c, d)` allows us to then wrap around it and accept `Tuple`s, stacked `Array`s, whatever we want, so it is perfectly OK to have it like this.

mfalt:
>  I would have tried going for the "double interface" accepting either the Tuple object (x,y) as input (to prox! for example) or the "stacked" Vector object [x; y], rather than the 5-arguments prox!(x, y, f, c, d). But maybe we can add this in the future.

The reason for going with the 5arg instead of stacked form is also that the stacked form can always be converted (using views) to the 5arg, the other way is sometimes not possible. The tuple and the 5arg is essentially equivalent though.

> We already tried tuple case with @mfalt.

Yes, but if @lostella thinks its a useful wrapper, then I have no objections to adding this wrapper too.

mi-volodin:
Okay, I get the tuples back and add the calls to `test_calls.jl` with tuple arguments - right? Or better put it in `test_graph`?

lostella:
I would say put them in both

mi-volodin:
Okay :)

@mfalt any additional suggestions? 
Also, see [this post](https://github.com/kul-forbes/ProximalOperators.jl/pull/40#discussion_r133940925).

lostella:
Yeah, don't worry too much about tests organization now, we will probably open a thread for that very soon and put those in order once and for (hopefully) all.

mi-volodin:
Alright, then here we go:
- [x] put `test_graph` to right testset
- [x] put `IndGraph` to docs file `functions.md`
- [x] add tuple-type methods back
- [x] include tests into `test_graph` with different calls

mi-volodin:
I am worried about one thing. `prox!` signatures are fine and not ambigous. 

However `prox_naive` are not so well.
`prox_naive(f, c, d)` and `prox_naive(f, cd, gamma)` has the same signatures. Thus, `gamma` becomes an obligatory argument here, to be able to handle the dispatching.

Everything else fixed, see checklist above.

EDIT: And test fails for IndAffine projection 😄 I also don't receive same point by IndAffine for Complex case.

EDIT2: Here the test case for Mac OS

```julia
using ProximalOperators

m, n = (50, 100)
t = 0
srand(t)

function runtest(m, n)
  A = rand(Complex{Float64}, m, n)
  x = randn(m + n) + im * randn(m + n)

  faff = IndAffine([A -eye(m)], zeros(Complex{Float64}, m))

  xa, fa = prox(faff, x)

  @test faff(xa) == 0.0
end

for i = 1:10000
  # srand(i) #if I uncomment this - everything works. I think it's just because it require larger matrices
  println(i)
  runtest(m, n)
end
```
On my machine it fails on `i = 16`.

P.S. Just to be sure made the same test for `IndGraph` - worked fine.

mi-volodin:
I can disable the test versus IndAffine now - it can be enabled later, when revision in PR#41 will be ready.

EDIT: As I see, the `[A -eye(m)] * x ~ 8e-14` which just do not pass tolerance in current version of indicator.

lostella (approved):

src/functions/indGraph.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    IndGraph(A)
+
+For matrix `A` (dense or sparse) returns the indicator function of the set
+```math
+S = \\{(x, y) : Ax = y\\}.
+```
+
+The evaluation of `prox!` uses direct methods based on LDLt (LL for dense cases) matrix factorization and backsolve.
+
```
lostella, line 14: I would here describe how the `Tuple`-friendly calls to function evaluation & `prox!` work: the whole library relies on the four-arguments `prox!`, the last argument `gamma` being optional. Therefore maybe it is better to describe the function in terms of single input/output objects. Something like

```
The `prox!` method operates on pairs `(x, y)` as input/output. So if `f = IndGraph(A)`,
while `(x, y)` and `(c, d)` are pairs of vectors of the same sizes, then `prox!((c, d), f, (x, y))`
writes to `(c, d)` the projection onto `S` of `(x, y)`.
```

Also, no need to spend words on the optional `gamma` below.

src/functions/indGraph.jl
```diff
@@ -0,0 +1,98 @@
+export IndGraph
+
+"""
+**Indicator of an operator graph**
```
lostella, line 4: I would also change the title of the docstring to
```
**Indicator of the graph of a linear operator**
```

lostella (approved):
These two very minor changes, then I guess we can merge!

lostella:
Actually I think we can merge now and then fix on `master` directly. Good job!

mi-volodin:
Wow! Thanks!
Actually my laptop has gone out of service suddenly, that is why I didn't do a thing about this PR.

PR #41, opened 2017-07-21
Improving IndAffine, LeastSquares, Quadratic

lostella:
In light of the recent discussions, I took a look to `IndAffine` to find that this type should probably be improved.

I’ve updated the **dense** case to perform as many operations as possible in-place. This resulted in performance improvements, see following comments.

@mfalt: do you think it’s possible to do similar things in the **sparse** case? I understand that using this low-level linear algebra routines is not the easiest thing, especially for sparse matrices/factorizations. Documentation is sometimes lacking?

**EDIT**

In light of the similar implementation details involved in handling these types, this PR now contains a restructuring of the `IndAffine`, `LeastSquares` and `Quadratic` types, together with a tentative improvement in their performance (by e.g. making as many operations as possible in-place).

**TODO**

- [x] Make sure to have as much as possible done in-place in `IndAffine`, `LeastSquares` and `Quadratic`.


lostella:
When running the following

```julia
using ProximalOperators
using BenchmarkTools
srand(0)
# define nonempty affine subspace (A, b)
m, n = 50, 200
A = randn(m, n)
x0 = randn(n)
b = A*x0
# construct indicator function
f = IndAffine(A, b)
# time projections
x = randn(n)
y = similar(x)
@benchmark prox!(y, f, x)
```

**Previous results** (6d98bd8)

```
BenchmarkTools.Trial:
  memory estimate:  25.34 KiB
  allocs estimate:  11
  --------------
  minimum time:     21.464 μs (0.00% GC)
  median time:      22.799 μs (0.00% GC)
  mean time:        26.144 μs (5.31% GC)
  maximum time:     1.872 ms (93.83% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

**New results** (f94f86c, this PR)

```
BenchmarkTools.Trial:
  memory estimate:  48 bytes
  allocs estimate:  3
  --------------
  minimum time:     12.717 μs (0.00% GC)
  median time:      13.409 μs (0.00% GC)
  mean time:        14.195 μs (0.00% GC)
  maximum time:     282.460 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

src/functions/indAffine.jl
```diff
@@ -60,8 +61,12 @@ function (f::IndAffine){R<:Real, T <: RealOrComplex{R}}(x::AbstractArray{T,1})
...
 
 function prox!{R<:Real, T<:RealOrComplex{R}, M<:DenseArray, V<:AbstractArray{T,1}}(y::V, f::IndAffine{T,M,V,M}, x::V, gamma::R=one(R))
-  res = f.A*x - f.b
-  y[:] = x - f.A'*(f.R\(f.R'\res))
+  A_mul_B!(f.res, f.A, x)
+  f.res .= f.b .- f.res
+  LinAlg.LAPACK.trtrs!('U', 'C', 'N', f.R, f.res);
+  LinAlg.LAPACK.trtrs!('U', 'N', 'N', f.R, f.res);
+  Ac_mul_B!(y, f.A, f.res)
+  y .+= x
```
mfalt: I've actually had significant slowdowns in the code due to lines like this, it seems like the broadcast is not as good as it should. Could you try to run the benchmark with the following code instead:
`using SugarBLAS: @blas!`
line 65:
```julia
@blas! f.res *= -1.0
@blas! f.res += f.b
```
line 69:
```julia
@blas! y += x
```
At least in my CG code, the following line: `p .= r .+ β.*p` took more time due to all the broadcast than the matrix-vector multiplications... but it was negligible when i changed it to
```julia
@blas! p *= β
@blas! p += r
```

mfalt:
Great work by the way, nice improvements. I think there is room for improvement in the sparse case, especially `RES = SparseArrays.CHOLMOD.Dense(f.A*x - f.b)` and `y .= x .- f.A'RRres`. Some places that look like allocations are probably not however, I think Dense() will just wrap the data for example. 

We might want to consider having different concrete types for the sparse and dense case (since they may require different storage). I had a long discussion with @mi-volodin yesterday, it may be beneficial to have seperate concrete types for the dense/sparse and direct/iterative versions of IndAffine. I have implementation for iterative IndAffine im my package, and I think @mi-volodin will propose a couple of different options for the direct and dense case.

PS: my implementation of iterative IndAffine is actually implementation of the more general graph problem, but with a linear term: `f([x;z]) = q'x+i(Ax-βz==b)`, where β is 1 or -1. See: https://github.com/mfalt/FirstOrderSolvers.jl/blob/lsrewrite/src/utilities/affinepluslinear.jl

lostella:
I agree, this will also lighten the type parameters significantly. I guess something like

```julia
abstract type IndAffine <: ProximableFunction end
struct IndAffineDense <: IndAffine [...] end
struct IndAffineSparse <: IndAffine [...] end
struct IndAffineIterative <: IndAffine [...] end
```

And similarly for `Quadratic` and `LeastSquares`, for example. This has also the advantage that all predicates `is_smooth`, `is_quadratic` etc can be defined over the `abstract type` once and for all.

mi-volodin:
Nice idea!

This is what we have already suggested in my `IndGraph` project.
But in a manner
```julia
...
struct IndGraphFat <: IndGraph [...] end
struct IndGraphSkinny <: IndGraph [...] end
```
Because efficient methods for `(m<n)` and `(m>n)` are different. 

Edit: I meant the replacement for IndAffineDense

lostella:
Ok, great! As for the suggestion of @mfalt regarding the `@blas!` macro: I'll give it a try, but let's see if it's possible to do this without increasing package dependencies. I'm already quite satisfied by the improvement in time/allocations (especially allocations).

mfalt:
The macro is just a short and easy command to call the built in functions, you can use macroexpand to see what is it's actually doing to avoid any dependencies, it's just calling axpy! I'm guessing.

lostella:
I have updated the PR description above: the edit mainly refers to the latest commit 3d0cd80. This organization of `IndAffine`, `LeastSquares`, `Quadratic` makes very much sense to me: for the moment the only necessary "type-splits" I saw were between the "direct" and "iterative" case. A further split could make two separate types for the "direct-dense" and "direct-sparse", but I would avoid this and rely on dispatch, unless we see it's necessary.

src/functions/indAffineDirect.jl
```diff
@@ -0,0 +1,105 @@
+### CONCRETE TYPE: DIRECT PROX EVALUATION
+# prox! is computed using a QR factorization of A'.
+
+struct IndAffineDirect{R <: Real, T <: RealOrComplex{R}, M <: AbstractMatrix{T}, V <: AbstractVector{T}, F <: Factorization} <: IndAffine
```
mi-volodin, line 4: Am I right If I say that it will fail to create an object on input
```julia
A = rand(m, n) + im * rand(m, n)
b = zeros(m)
``` 
for any arbitrary `m,n` that is acceptable. 

I think it's not okay, but maybe there's a reasonable workaround without adding `T2`.

Testcase:
```julia
const RorC = Union{Real, Complex}

function testt(a::T, b::T) where {T<: RorC}
  a + b
end

testt(2., 3.)
testt(2. + im * 3., 3. + im * 2.)
testt(2. + im * 3., 3.) #Fails
```

mfalt: I think that you are right that it will fail to create a type. There should be ways to get around this by adding more type parameters.
It is easy to see that we need `res::promote_type(typeof(A), typeof(b))`, the type of the `b` we store is a bit trickier since we use norm of A to scale, with for example `b::Array{Float32,1}`, `A::Array{Complex{Float64},2}` we have to store `b::Array{Float64,1}`. This means that we need one type for `A`, one for `b`, and one for `res`, and I don't see that we get any significant improvement compared to doing something like
```julia
function IndAffineDirect{T<:RealOrComplex{<:Real}}(A::AbstratMatrix{T} b::AbstractVector)
  b2 = similar(b, T)
  b2 .= b
  IndAffineDirect(A, b2)
end
```
but then we run into problems when `b` is of "higher accuracy" than `A`. I'm not sure if there is an elegant solution without having one type for each value. But I think that if we have too many type parameters somewhere we risk that julia stops specializing on them.
The only case I can see where we would benefit from having different types is if `A` in some really low accuracy format, like `Array{Int8,2}`, but since the factorization only works with Float32, Float64, and the corresponding Complex, we won't get any real benefit.

mi-volodin: In fact all this could rely on caller (meaning user). It just would be fine to leave a note about it in the docs 😄  Just because julia errors about the type mismatch in this case are not very informative.

I solved my problem by extracting `T` from `A` and create zeros with this type.

lostella:
I couldn't do anything more (for the moment) than increasing the tolerance for evaluating `f::IndAffine` in order to overcome the issue highlighted by @mi-volodin in #40.

@mfalt: it looks like in fact the `shift` option is **not** available for dense Cholesky factorizations. This is really unfortunate.

If nobody thinks there are big issues here, I would consider merging this PR. I decided to remove other goals from this PR to then open appropriate issues: smaller PRs are faster to merge, and it's easier to reason on smaller individual problems.

mfalt:
I think it's ready to merge.

PR #42, opened 2017-08-26
Further improvements of iterative versions of prox!

lostella:
In this PR I switch to the IterativeSolvers package for the `cg!` methods used in `QuadraticIterative` and `LeastSquaresIterative`. Furthermore, an iterative version of `IndAffine` is implemented using alternating projections.

Open questions:
* How should we allow for the passing of options to the internally used iterative solvers? (Like `IterativeSolvers.cg!` in this case)
* `IterativeSolvers.cg!` considers the value of the given vector (where the solution is written) as initial iterate. However, in `LeastSquaresIterative` there is the “fat” case where we resort to Woodbury’s identity to reduce the problem size. How to allow custom initialization of `IterativeSolvers.cg!` in that case?

**TODO**
- [x] Add tests for `IndAffine(A, b; iterative=true)`


lostella:
@mfalt you're the expert on projections here :-) What do you think of the `IndAffineIterative`?

mfalt:
I've been thinking a bit about the IndAffineIterative, and I don't think the Alternating Projections is very efficient, I think something like this method could be state of the art: https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/src/craigmr.jl
I have my own implementation that doesn't use any packages too.
The other iterative solvers you added looks good.
Edit:
I am not sure what the best way to warmstart is, I have implemented a warm-start for the CRAIG-MR, but I am not sure how it compares to, for example, CG on the normal equations `(AA' + I) y = b`.

lostella:
I think we can merge this, to then later update the behaviour of `IndAffineIterative` (I'm not familiar with those methods, and how to use them to project a specific point; also, we probably need to use a custom implementation since Krylov.jl is unregistered?)

Issue #43, opened 2017-10-20
Gradients


mfalt:
We seem to be missing most of the implementations of `gradient`/`gradient!`. Was this intentional or should I start adding them?
Should we define gradients even for functions that are not differentiable everywhere by returning one of the subgradients?
For example `NormL1` could return simply `0` vector at `0`.

It would be nice with a short command like
`∇(f,x) = gradient(f,x)` and
`∇!(y,f,x) = gradient!(y,f,x)`
any objections to me adding that @lostella ?
One could also use `∂(f,x)` to signify that its part of the subdifferential, but since we would only return a set, and not a point, that could be confusing as well. 

Completed gradients (in PRs) out of non-Indicator functions that are either: convex (i.e. suggradients exist) or differentiable
- [x] ElasticNet
- [x] NormL1
- [x] NormL2
- [ ] NormL21
- [x] NormLinf
- [ ] NuclearNorm
- [x] SqrNormL2 (existed previously)
- [x] HingeLoss (Convex!?)
- [x] HuberLoss (existed previously)
- [x] LeastSquares (existed previously)
- [x] Linear (fixed)
- [x] LogBarrier
- [ ] Maximum (Using sum largest, convex?, differentiable?)
- [x] Quadratic (existed previously)
- [x] QuadraticIterative (existed previously)
- [x] SumPositive
- [x] DistL2
- [x] SqrDisrL2

lostella:
Hey @mfalt: all the `gradient!` operations are there, just lazily implemented :-p

Jokes apart, you can surely implement the missing ones if you want, and I also completely agree with returning one subgradient for nondifferentiable functions (why not?). We should coherently update the docs in this case.

As per the shortcut: sure, I'm not a fan of using weird unicode characters in code, but that's a non-breaking additional feature.

mfalt:
Great, I'll start implementing ASAP, I wanted to do a presentation for some colleagues, and realized that forward-backward was not as straight forward (no pun intended) to implement as I had hoped because of the missing gradients.
Btw, have you seen my gitter message @lostella ?

lostella:
No I haven't opened it in a while, I'll go look at it asap

panpat:
Guys, instead of reinventing the wheel what about letting JuliaDif compute the gradients using AD?

lostella:
True, I’m wondering what would that yield in case of nonsmooth functions. It is worth checking for sure.

panpat:
[...]

PR #44, opened 2017-10-23
Gradients2

mfalt:
Implemented most of the gradients according to issue #43, test still missing.
Update: Tests added, still missing: NormL21, Maximum, NuclearNorm, but those can be added at a later date if needed.

### DIFF
--- a/docs/src/functions.md
+++ b/docs/src/functions.md
@@ -75,3 +75,9 @@ Maximum
 Quadratic
 SumPositive
 ```
+
+## Distances
+```@docs
+DistL2
+SqrDistL2
+```
--- a/src/functions/indSOC.jl
+++ b/src/functions/indSOC.jl
@@ -27,7 +27,7 @@ is_convex(f::IndSOC) = true
 is_set(f::IndSOC) = true
 
 function prox!{T <: Real}(y::AbstractArray{T,1}, f::IndSOC, x::AbstractArray{T,1}, gamma::T=one(T))
-  nx = norm(x[2:end])
+  @views nx = norm(x[2:end])
   t = x[1]
   if t <= -nx
     y[:] = 0.0
@@ -36,7 +36,7 @@ function prox!{T <: Real}(y::AbstractArray{T,1}, f::IndSOC, x::AbstractArray{T,1
   else
     r = 0.5 * (1 + t / nx)
     y[1] = r * nx
-    y[2:end] = r * x[2:end]
+    @views y[2:end] .= r .* x[2:end]
   end
   return 0.0
 end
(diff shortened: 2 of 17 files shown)


docs/src/functions.md
```diff
@@ -75,3 +75,9 @@ Maximum
 Quadratic
 SumPositive
 ```
+
+## Distances
```
lostella, line 79: This I have also put in the section describing calculus rules, since these are really modifiers of some other (indicator) functions; but I guess they can stay in both places.

src/functions/indSOC.jl
```diff
@@ -27,7 +27,7 @@ is_convex(f::IndSOC) = true
 is_set(f::IndSOC) = true
 
 function prox!{T <: Real}(y::AbstractArray{T,1}, f::IndSOC, x::AbstractArray{T,1}, gamma::T=one(T))
-  nx = norm(x[2:end])
+  @views nx = norm(x[2:end])
```
lostella, line 30: Nice, I didn’t know this!

src/functions/linear.jl
```diff
@@ -17,6 +17,9 @@ end
 
 Linear(c::A) where {R <: RealOrComplex, A <: AbstractArray{R}} = Linear{R, A}(c)
 
+is_separable(f::Linear) = true
+is_convex(f::Linear) = true
```
lostella, line 21: I would also add `is_smooth = true` if not already there.

src/functions/quadratic.jl
```diff
@@ -19,6 +19,7 @@ If `iterative=true`, then `prox!` is evaluated approximately using an iterative
 
 abstract type Quadratic <: ProximableFunction end
 
+is_convex(f::Quadratic) = true
```
lostella, line 22: This I have not previously put because one should really check that `Q` is positive semidefinite...

src/functions/linear.jl
```diff
@@ -28,7 +31,6 @@ function prox!(y::AbstractArray{RC}, f::Linear{RC, A}, x::AbstractArray{RC}, gam
 end
```
mfalt, line 31: Not changed in PR, but should be
`y .= x .- gamma.*(f.c)`
for efficiency.

lostella (approved):

src/functions/quadratic.jl
lostella: Any thoughts on this @mfalt ? Of course checking the eigenvalues is not viable, but maybe the user should be able to assert whether the function is convex or not?

lostella: No ok, the docstring says we assume `Q` to be positive definite (should be semidefinite, to be corrected). We may want to change this in the near future, not all quadratics are convex!

lostella (approved):

src/functions/quadratic.jl
mfalt: Yes, I added this because of the documentation, but you are right, maybe some keyword from the user, acknowledging if it is positive(semi)-definite would be good.

docs/src/functions.md
mfalt: They were in the FUNCTIONS.md in the root, so I copied them here. But you are right, we should think of how we can unify the different documentations.

PR #45, opened 2017-11-08
fixed small bug on SlicedSeparableSum

nantonel:
Small bug on `SlicedSeparableSum` when `idxs` was composed by `Array{Int,1}`. 
Apparently
```julia
julia> x = collect(1:10);

julia> typeof(x) <: AbstractArray{Integer,1}
false
```
changing `AbstractArray{Integer,1}` to `AbstractArray{Int,1}` fixed the problem.

### DIFF
--- a/src/calculus/slicedSeparableSum.jl
+++ b/src/calculus/slicedSeparableSum.jl
@@ -28,7 +28,7 @@ struct SlicedSeparableSum{S <: Tuple, T <: AbstractArray, N} <: ProximableFuncti
 end
 
 function SlicedSeparableSum(fs::S, idxs::T) where {N,
-						   S <: NTuple{N,Union{subtypes(ProximableFunction)...}},
+						   S <: Tuple{Vararg{<:ProximableFunction,N}},
 						   M, 
 						   I <: Integer, 
 						   T1 <: NTuple{M,Union{I,
--- a/src/calculus/slicedSeparableSum.jl
+++ b/src/calculus/slicedSeparableSum.jl
@@ -20,21 +20,25 @@ g(x) = ∑_\{i=1\}^k f(x_{J_i}).
 ```
 """
 
-immutable SlicedSeparableSum{S <: Tuple, T <: AbstractArray, N} <: ProximableFunction
+struct SlicedSeparableSum{S <: Tuple, T <: AbstractArray, N} <: ProximableFunction
   fs::S    # Tuple, where each element is a Vector with elements of the same type; the functions to prox on
   # Example: S = Tuple{Array{ProximalOperators.NormL1{Float64},1}, Array{ProximalOperators.NormL2{Float64},1}}
   idxs::T  # Vector, where each element is a Vector contining the indices to prox on
   # Example: T = Array{Array{Tuple{Colon,UnitRange{Int64}},1},1}
 end
 
-function SlicedSeparableSum{S <: AbstractArray, T <: AbstractArray}(fs::S, idxs::T)
-  if size(fs) != size(idxs)
-    error("size(fs) must coincide with size(idxs)")
-  else
-    for k in eachindex(idxs)
-      cond = [typeof(t) <: Integer || typeof(t) <: AbstractArray{Int,1} || typeof(t) <: Colon || typeof(t) <: Range for t in idxs[k]]
-			!all(cond) ? error("invalid index $(k)") : nothing
-    end
+function SlicedSeparableSum(fs::S, idxs::T) where {N,
+						   S <: NTuple{N,Union{subtypes(ProximableFunction)...}},
+						   M, 
+						   I <: Integer, 
+						   T1 <: NTuple{M,Union{I,
+								       AbstractArray{I},
+								       Colon,
+								       Range
+								       }
+							       },
+						   T <:NTuple{N,T1}
+						   }
     ftypes = DataType[]
     fsarr = Array{Any,1}[]
     indarr = Array{eltype(idxs),1}[]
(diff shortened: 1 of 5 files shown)


src/calculus/slicedSeparableSum.jl
```diff
@@ -32,7 +32,7 @@ function SlicedSeparableSum{S <: AbstractArray, T <: AbstractArray}(fs::S, idxs:
     error("size(fs) must coincide with size(idxs)")
   else
     for k in eachindex(idxs)
-      cond = [typeof(t) <: Integer || typeof(t) <: AbstractArray{Integer,1} || typeof(t) <: Colon || typeof(t) <: Range for t in idxs[k]]
+      cond = [typeof(t) <: Integer || typeof(t) <: AbstractArray{Int,1} || typeof(t) <: Colon || typeof(t) <: Range for t in idxs[k]]
```
lostella: The problem was that in Julia `S <: T` does not imply `AbstractArray{S} <: AbstractArray{T}`.

I guess this whole check should somehow go in the type parameters constraints (line 30). However, to keep it like this for now, it's probably better to have something like
```
typeof(t) <: AbstractArray && eltype(t) <: Integer
```

lostella (changes requested):
See my suggestion, then we can merge (but those type checks we will need to adjust at some point)

src/calculus/slicedSeparableSum.jl
lostella: The reason being that `Int <: Integer`

nantonel: if we want to do these checks directly on the constructor (and I am in favour of this) I think it would be better to use `Tuples` instead of `Arrays` both for `idxs` and `fs`! 

That is because 
```julia 
julia> typeof([1,:])
Array{Any,1}
```
while 
```julia
julia> typeof((1,:))
Tuple{Int64,Colon} 
```
which has the advantage being capable of doing this:
```julia
julia> typeof((1,:)) <: NTuple{2,Union{Integer,Colon}}
true
```

nantonel: I'll try to do this, see if you like it ;)

lostella: You know better how `SlicedSeparableSum` works, you have carta bianca ;-)

nantonel:
I fixed the type checks! but I had to change the call which now uses `Tuples` instead of `Array`. I think this is even more consistent with `SeparableSum`...

mfalt:
Could you give me a few days to consider and test the implications of this change before we merge. I am quite dependent on the `SlicedSeperatbleSum` in my packages.
The only implications I can think of at the top of my head is that tuples are non-mutable and can't be constructed iteratively.

lostella:
from my side: go ahead, I added you as a reviewer

src/calculus/slicedSeparableSum.jl
```diff
@@ -20,21 +20,25 @@ g(x) = ∑_\{i=1\}^k f(x_{J_i}).
...
-function SlicedSeparableSum{S <: AbstractArray, T <: AbstractArray}(fs::S, idxs::T)
-  if size(fs) != size(idxs)
-    error("size(fs) must coincide with size(idxs)")
-  else
-    for k in eachindex(idxs)
-      cond = [typeof(t) <: Integer || typeof(t) <: AbstractArray{Integer,1} || typeof(t) <: Colon || typeof(t) <: Range for t in idxs[k]]
-			!all(cond) ? error("invalid index $(k)") : nothing
-    end
+function SlicedSeparableSum(fs::S, idxs::T) where {N,
+						   S <: NTuple{N,Union{subtypes(ProximableFunction)...}},
```
mfalt: Does 'NTuple{N,Union{subtypes(ProximableFunction)...}}' work properly for types defined after loading this module? I.e will this break code that defines new subtypes?
'Tuple{Vararg{<:ProximableFunction,N}})' should work.

src/calculus/slicedSeparableSum.jl
```diff
@@ -20,21 +20,25 @@ g(x) = ∑_\{i=1\}^k f(x_{J_i}).
...
+						   M, 
+						   I <: Integer, 
+						   T1 <: NTuple{M,Union{I,
```
mfalt: This seems like a change from the previous behavior, I am not sure which way is correct. I think this forces all elements in `T` to be of same length, and have a common type `I`. For example `idxs=((1,:), (Int32(2),:))`, wouldn't work, neither would `idx=((1,1,:),(1,2,:),(2,:))` although this might be good. I think that we should at least consider
```
T1 <: NTuple{M,Union{I,
                     AbstractArray{I},
                     Colon,
                     Range
                     } where {I<:Integer}
            }
```
and if we can motivate different lengths:
```
T1 <: NTuple{M,Union{I,
                     AbstractArray{I},
                     Colon,
                     Range
                     } where {I<:Integer}
             } where {M}
```

nantonel:
@mfalt sure thing, there is no hurry! :smile:

mfalt (changes requested):
After the proposed changes/considerations, I see no reason not to pull this. I may have to update my code that uses this feature, but I think that might be best.

src/calculus/slicedSeparableSum.jl
nantonel: thanks I didn't know you could use that notation!

src/calculus/slicedSeparableSum.jl
nantonel: well I prefer to have the same length... having different ones seems to be more prone to errors given that we currently don't check if there are repeated indexes.

nantonel:
@mfalt let me know if you agree with the latest changes! In that case feel free to Merge. Thanks for your suggestions!

PR #46, opened 2017-11-26
Added logistic, squared hinge, cross entropy loss functions

lostella:
Title is self explanatory. Only `gradient!` is provided though, we may add `prox!` using a very efficiently implemented Newton method.


nantonel:
I have some other function for which I only implemented the gradient. Namely a smooth version of the Hinge loss and a Cross Entropy. Maybe I could add these to this PR, what do you think?

nantonel:
also, I'm wondering, for these functions that currently don't have `prox!` we could add a predicate `is_proximable(...)= false`?

lostella:
Yes, you could append them to this PR. But I would rather not add a predicate `is_proximable`, and instead push to have all `prox!` implementations eventually.

I say: add the `gradient!`-only functions here, and upon merging we open an issue listing with the missing `prox!` implementations.

BTW: one way to compute `prox!` for smooth functions by default would be to do it numerically e.g. using L-BFGS.

nantonel:
yes it would be nice to have such thing with L-BFGS for all smooth function without dedicated `prox!`

mfalt:
I agree that we should aim to have prox available for all functions, they are `<: ProximableFunction` after all.

PR #50, opened 2018-02-09
Sum gradient

currymj:
I hewed pretty closely to the existing code in making this. Please let me know if it isn't designed right or doesn't fit in with the rest of the code -- I'm happy to make modifications.

Some notable things:

I've assumed the user is going to pass in functions that all have the same domain. Is there a way to enforce this through the type system? Might it make sense to define a custom constructor to check?

I haven't defined any code to compute `prox`, and I've therefore set `is_prox_accurate` to be always false. The other properties I believe are all closed under addition, right?

I couldn't get away without allocating an array in `gradient!`, in order to keep track of the sum of each individual call.

I just have two simple tests; do things need to be more thorough?

### DIFF
--- a/src/calculus/sum.jl
+++ b/src/calculus/sum.jl
@@ -32,17 +32,17 @@ is_quadratic(f::Sum) = all(is_quadratic.(f.fs))
 is_generalized_quadratic(f::Sum) = all(is_generalized_quadratic.(f.fs))
 is_strongly_convex(f::Sum) = all(is_convex.(f.fs)) && any(is_strongly_convex.(f.fs))
 
-function (sumobj::Sum)(x)
-    sum = 0.0
+function (sumobj::Sum)(x::AbstractArray{T}) where {R, T <: Union{R, Complex{R}}}
+    sum = zero(R)
     for f in sumobj.fs
         sum += f(x)
     end
     sum
 end
 
-function gradient!(grad::T, sumobj::Sum, x::T) where T
+function gradient!(grad::AbstractArray{T}, sumobj::Sum, x::AbstractArray{T}) where {R, T <: Union{R, Complex{R}}}
     # gradient of sum is sum of gradients
-    val = 0.0
+    val = zero(R)
     # to keep track of this sum, i may not be able to
     # avoid allocating an array
     temp = zeros(grad)
--- a/src/calculus/sum.jl
+++ b/src/calculus/sum.jl
@@ -1,0 +1,61 @@
+export Sum
+
+"""
+**Sum of functions**
+
+    Sum(f₁,…,fₖ)
+
+Given functions `f₁` to `fₖ`, returns their sum
+
+```math
+g(x) = ∑_\{i=1\}^k f_i(x).
+```
+
+The functions `fᵢ` should all have the same domain, which will be the domain of
+the result. If all `fᵢ` have gradients then this function will also have a
+gradient.
+
+"""
+
+immutable Sum{T <: Tuple} <: ProximableFunction fs::T end
+
+Sum(fs::Vararg{ProximableFunction}) = Sum((fs...))
+
+# note: is_prox_accurate false because prox in general doesn't exist?
+is_prox_accurate(f::Sum) = false
+is_convex(f::Sum) = all(is_convex.(f.fs))
+is_set(f::Sum) = all(is_set.(f.fs))
+is_singleton(f::Sum) = all(is_singleton.(f.fs))
+is_cone(f::Sum) = all(is_cone.(f.fs))
+is_affine(f::Sum) = all(is_affine.(f.fs))
+is_smooth(f::Sum) = all(is_smooth.(f.fs))
+is_quadratic(f::Sum) = all(is_quadratic.(f.fs))
+is_generalized_quadratic(f::Sum) = all(is_generalized_quadratic.(f.fs))
+is_strongly_convex(f::Sum) = all(is_strongly_convex.(f.fs))
+
+function (sumobj::Sum)(x)
+    sum = 0.0
+    for f in sumobj.fs
+        sum += f(x)
+    end
+    sum
+end
+
+function gradient!(grad, sumobj::Sum, x)
+    # gradient of sum is sum of gradients
+
+    val = 0.0
+    # to keep track of this sum, i may not be able to
+    # avoid allocating an array
+    temp = zeros(grad)
+    for f in sumobj.fs
+        val += gradient!(temp, f, x)
+        grad += temp
+    end
+    return val
+end
+
+fun_name(f::Sum) = "sum"
+fun_dom(f::Sum) = fun_dom(f.fs[1]) # check to make sure same??
+fun_expr(f::Sum) = "x ↦ f₁(x) + … + fₖ(x)"
+fun_params(f::Sum) = "n/a"
--- a/src/calculus/sum.jl
+++ b/src/calculus/sum.jl
@@ -25,13 +25,12 @@ Sum(fs::Vararg{ProximableFunction}) = Sum((fs...))
 is_prox_accurate(f::Sum) = false
 is_convex(f::Sum) = all(is_convex.(f.fs))
 is_set(f::Sum) = all(is_set.(f.fs))
-is_singleton(f::Sum) = all(is_singleton.(f.fs))
 is_cone(f::Sum) = all(is_cone.(f.fs))
 is_affine(f::Sum) = all(is_affine.(f.fs))
 is_smooth(f::Sum) = all(is_smooth.(f.fs))
 is_quadratic(f::Sum) = all(is_quadratic.(f.fs))
 is_generalized_quadratic(f::Sum) = all(is_generalized_quadratic.(f.fs))
-is_strongly_convex(f::Sum) = all(is_strongly_convex.(f.fs))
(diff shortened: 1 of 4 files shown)


src/calculus/sum.jl
```diff
@@ -0,0 +1,60 @@
...
+
+immutable Sum{T <: Tuple} <: ProximableFunction fs::T end
+
+Sum(fs::Vararg{ProximableFunction}) = Sum((fs...))
+
+# note: is_prox_accurate false because prox in general doesn't exist?
+is_prox_accurate(f::Sum) = false
+is_convex(f::Sum) = all(is_convex.(f.fs))
+is_set(f::Sum) = all(is_set.(f.fs))
+is_singleton(f::Sum) = all(is_singleton.(f.fs))
```
lostella: This I would avoid defining

src/calculus/sum.jl
```diff
@@ -0,0 +1,60 @@
...
+is_cone(f::Sum) = all(is_cone.(f.fs))
+is_affine(f::Sum) = all(is_affine.(f.fs))
+is_smooth(f::Sum) = all(is_smooth.(f.fs))
+is_quadratic(f::Sum) = all(is_quadratic.(f.fs))
+is_generalized_quadratic(f::Sum) = all(is_generalized_quadratic.(f.fs))
+is_strongly_convex(f::Sum) = all(is_strongly_convex.(f.fs))
```
lostella: I would say that f is strongly convex if it is convex and at least one of its summands is strongly convex, which is more general than this if I’m not wrong

src/calculus/sum.jl
lostella: I mean: the sum of set indicators is the indicator of the intersection; the intersection of singletons is a singleton if the sets are all equal, which is a nasty condition to check I guess, so I would leave this undefined (and let it default to False).

mfalt:
I have a few comments, but I am not sure if we should care about all of them by now.

I am pretty sure `for f in sumobj.fs` is type unstable since `f` can change type, which could significantly reduce the speed if each `f(x)` is cheap to compute. I am not sure if there is a nice way to solve this in julia or if we can expect new optimizations that will hande this. I have found that the following code can work well on other problems: (i was unable to test the code right now, so there is probably some mistakes in it)
```julia
customreduce(fs::Tuple{Vararg{<:ProximableFunction},N}, x) where N = fs[1](x) + customreduce(tail(fs), x)
customreduce(fs::Tuple{<:ProximableFunction,}, x) = fs[1](x)
```
Another alternative is to do loop unrolling, either manually using `@generated` which can be quite ugly, or using Unrolled.jl which introduces another dependency.

The gradient could similarly be done
```julia
function customreduce!(grad, temp, fs::Tuple{Vararg{<:ProximableFunction},N}, x) where N
  val = gradient!(temp, fs[1], x)
  grad .+= temp
  return val + customreduce(grad, tmp, tail(fs), x)
end
function customreduce!(grad, temp, fs::Tuple{<:ProximableFunction,}, x)
  val = gradient!(temp, fs[1], x)
  grad .+= temp
  return val
end
```
The second functions could instead match `fs::Tuple{}` if we can figure out what type `val` should be in the no argument case. For example
```julia
customreduce!(grad, temp, fs::Tuple{}, x) = zero(SomeType)
```
I think you are missing something like `grad .= zero(eltype(T))`, and it should suffice to make `tmp = similar(grad)`
And maybe `val` shouldnt default to `0.0`?

src/calculus/sum.jl
```diff
@@ -0,0 +1,60 @@
...
+
+function gradient!{T}(grad::T, sumobj::Sum, x::T)
+    # gradient of sum is sum of gradients
+    val = 0.0
+    # to keep track of this sum, i may not be able to
+    # avoid allocating an array
+    temp = zeros(grad)
+    for f in sumobj.fs
+        val += gradient!(temp, f, x)
+        grad[:] += temp
```
lostella: Maybe here `grad .+= temp` is more efficient: this `[:]` notation to make in-place operations may be do unnecessary allocations, you can check that using @time or BenchmarkTools.

src/calculus/sum.jl
```diff
@@ -0,0 +1,60 @@
...
+
+function (sumobj::Sum)(x)
+    sum = 0.0
+    for f in sumobj.fs
+        sum += f(x)
+    end
+    sum
+end
+
+function gradient!{T}(grad::T, sumobj::Sum, x::T)
```
lostella: Here you can use the `where` notation:

```
function gradient!(...) where {T}
...
end
```

lostella (changes requested):

src/calculus/sum.jl
lostella: I mean using `@time`, I’m afraid I inadvertently involved someone in the conversation :-)

currymj: it seems like they are equivalent, but i'd rather lean on the dot notation just because it looks nicer.

currymj:
val defaulting to 0.0 i just took from `SeparableSum`, but it does seem worrisome.

ideally I'd like to be able to do `zero(T)`, where `T` is somehow determined by looking at the return types of elements of `sumobj.fs` (which must all be the same, although the code doesn't enforce this currently).

if there's a way to encode that domain and codomain must be the same in the type system somehow, that seems like it would also solve the type instability problem for looping over the constituent functions.

src/calculus/sum.jl
```diff
@@ -0,0 +1,59 @@
...
+is_set(f::Sum) = all(is_set.(f.fs))
+is_cone(f::Sum) = all(is_cone.(f.fs))
+is_affine(f::Sum) = all(is_affine.(f.fs))
+is_smooth(f::Sum) = all(is_smooth.(f.fs))
+is_quadratic(f::Sum) = all(is_quadratic.(f.fs))
+is_generalized_quadratic(f::Sum) = all(is_generalized_quadratic.(f.fs))
+is_strongly_convex(f::Sum) = all(is_convex.(f.fs)) && any(is_strongly_convex.(f.fs))
+
+function (sumobj::Sum)(x)
+    sum = 0.0
```
lostella: To answer your question, here you could be doing 
```
function (sumobj::Sum)(x::AbstractArray{T}) where {R, T <: Union{R, Complex{R}}}
    sum = zero(R)
    ...
```

And similarly in `gradient!`.

This should be type stable provided that all component functions satisfy the same assumption, i.e. if they evaluate to R where R is the underlying real type of their argument. I’m not sure that *all* the functions implemented satisfy this, but if they don’t then an issue should be opened and this should be fixed.

mfalt: This looks a bit weird, if `T=Complex{Float64}`, then can't `R = Complex{Float64} <: Union{Complex{Float64}, Complex{Complex{Float64}}}`? I'm not sure how `T` gets decided here, Maybe we should at least say `R<:Real`?

lostella: Yes, of course :-)

src/calculus/sum.jl
```diff
@@ -0,0 +1,59 @@
...
+
+function (sumobj::Sum)(x::AbstractArray{T}) where {R, T <: Union{R, Complex{R}}}
+    sum = zero(R)
+    for f in sumobj.fs
+        sum += f(x)
+    end
+    sum
+end
+
+function gradient!(grad::AbstractArray{T}, sumobj::Sum, x::AbstractArray{T}) where {R, T <: Union{R, Complex{R}}}
```
lostella, line 43: As pointed out by @mfalt, here it is safer to restrict `R <: Real`, otherwise when `x::Array{Complex{Float64}}` is passed, it may be that both `R` and `T` get to be `Complex{Float64}`.

src/calculus/sum.jl
```diff
@@ -0,0 +1,59 @@
...
+    # gradient of sum is sum of gradients
+    val = zero(R)
+    # to keep track of this sum, i may not be able to
+    # avoid allocating an array
+    temp = zeros(grad)
```
lostella:
```
grad .= zero(R)
temp = similar(grad)
```

test/test_sum.jl
```diff
@@ -0,0 +1,15 @@
...
+f = Sum(f1, f2)
+xtest = randn(10)
+
+result = f1(xtest) + f2(xtest)
+@test f(xtest) ≈ result
+
+grad1, val1 = gradient(f1, xtest)
+grad2, val2 = gradient(f2, xtest)
+
+gradsum, valsum = gradient(f, xtest)
```
lostella, line 14: Here I would use `gradient!`, passing as result vector `gradsum` which is initialized to something nonzero (to test the fix in my comment above).

lostella (changes requested):
Some fixes are required, then we can merge.

src/calculus/sum.jl
lostella, line 48:
```
grad .= zero(T)
temp = similar(grad)
```

src/calculus/sum.jl
lostella: And the same applies to the definition above (the function evaluation)

lostella:
Hi, are there any news on this? This really requires minor changes (see comments above) then can be merged 👍

lostella:
Closing, this was re-issued in #55 with required final adjustments.

Issue #52, opened 2018-03-09
Projection on polyhedral set


panpat:
It would be nice to have it there. I mean for general polyhedral sets
lb<=A*x<=ub
xmin<=x<=xmax

It could either be implemented by calling an external QP solver or solve the dual using e.g., PANOC.
Also there should be an option to warm-start the whole thing.

lostella:
This could this be an instance of the more general case g(x) = f(Ax), for convex, proximable f, I guess.

panpat:
Even better, although you wouldn't be able to use a QP solver for the prox of f(Ax), only dual PANOC (NAMA). But if it is a QP, then we should be able to have the option to use something like GUROBI for small problems.

lostella:
PR #54 contains an implementation of this (no warm-start yet).

mfalt:
I just submitted a paper for a QP solver. It is supposed to be really efficent projecting on Ax=b, Cx<=d, when the set of constraints is relatively small, by solving the dual problem.
The algorithm, completely in julia is found here: https://github.com/mfalt/QPDAS.jl
We could add it as an option for "solver".
The interface is almost the same, we would have to rewrite when lower bound is equal to upper bound to an equality, and there is no added efficiency for using xmin/xmax instead of lb/ub.
And it allows for any of these bounds to be changed without re-factorizing, and warmstart is done automatically.

It should beat OSQP pretty comfortably unless the number of inequalites is large.

PR #54, opened 2018-04-13
Adding indicator of polyhedral set

lostella:
This adds the indicator of a polyhedral set
```
{x : xmin <= x <= xmax, l <= Ax <= u}
```
For the moment, the projection is only implemented using [OSQP](https://github.com/oxfordcontrol/OSQP.jl). This PR is backward compatible in that it just adds a function. The only additional change is a minor modification to a test routine.


lostella:
Relevant for #52

PR #56, opened 2018-06-26
Add simd to for loops

baggepinnen:
A simple benchmark with `julia -O 3` shows almost double performance in some cases

Example, `prox2!` is the updated prox function with `@simd` added. Run on julia v0.6.2
```julia
using ProximalOperators
using BenchmarkTools

function prox2!{T <: Real, R <: Real}(y::AbstractArray{R}, f::NormL1{T}, x::AbstractArray{R}, gamma::Real=1.0)
  n1y = zero(R)
  gl = gamma*f.lambda
  @inbounds @simd for i in eachindex(x)
    y[i] = x[i] + (x[i] <= -gl ? gl : (x[i] >= gl ? -gl : -x[i]))
    n1y += y[i] > 0 ? y[i] : -y[i]
  end
  return f.lambda*n1y
end


f = NormL1(1.0)
x = randn(1_000_000);
y = similar(x);

julia> @benchmark prox!($y,$f,$x)

BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.866 ms (0.00% GC)
  median time:      2.897 ms (0.00% GC)
  mean time:        2.984 ms (0.00% GC)
  maximum time:     5.647 ms (0.00% GC)
  --------------
  samples:          1671
  evals/sample:     1

julia> @benchmark prox2!($y,$f,$x)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.456 ms (0.00% GC)
  median time:      1.479 ms (0.00% GC)
  mean time:        1.553 ms (0.00% GC)
  maximum time:     5.520 ms (0.00% GC)
  --------------
  samples:          3206
  evals/sample:     1
```

### DIFF
--- a/src/functions/normL1.jl
+++ b/src/functions/normL1.jl
@@ -60,6 +60,7 @@ function (f::NormL1{A}){A <: AbstractArray}(x::AbstractArray)
 end
 
 function prox!{A <: AbstractArray, R <: Real}(y::AbstractArray{R}, f::NormL1{A}, x::AbstractArray{R}, gamma::Real=1.0)
+  @assert length(y) == length(x) == length(f.lambda)
   fy = zero(R)
   @inbounds @simd for i in eachindex(x)
     gl = gamma*f.lambda[i]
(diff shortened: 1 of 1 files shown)


nantonel:
Hi, 

this is nice, however isn't it a little bit dangerous in general to use `@inbouds`? 

> Using `@inbounds` may return incorrect results/crashes/corruption for out-of-bounds indices. The user is responsible for checking it manually. Only use `@inbounds` when it is certain from the information locally available that all accesses are in bounds.

Maybe we could do a check like `size(y) == size(x)`?

baggepinnen:
I agree, I added the appropriate checks!

Issue #47, opened 2017-11-27
Compatibility with 0.7/1.0


lostella:
Currently tests fail on nightly. Since feature freeze for Julia 1.0 is about to hit, it will be soon time to start making the package forward-compatible. We can use this space to list criticalities and discussions on that regards.

**Update - 25/03/2018**

I'm currently going through this. General things found so far:
- [x] the `Base.LinAlg` module is now `LinearAlgebra`
- [x] `Base.A_mul_B!` has become `LinearAlgebra.mul!`
- [x] `Base.Ac_mul_B!(y, A, x)` is now `LinearAlgebra.mul!(y, adjoint(A), x)`
- [x] sparse factorizations moved to the `SuiteSparse` module (from `Base.SparseArrays`)
- [x] random number generation functions moved to the `Random` module (from `Base`)
- [x] `@printf` moved to the `Printf` module (from `Base`)
- [x] warnings are generated when `.+` (and similarly for other operators) is not appropriately spaced away from the operands (example: `2.+x` confuses the parser on the nature of first summand `2` or `2.0` I guess)
- [x] when broadcast was not explicitly used, e.g. `x::AbstractArray + y::Number`, now we get a warning
- [x] `zeros(a::AbstractArray)` is deprecated in favor of `zero`, `fill`, `fill!`
- [x] the old form `function f{T}(x::T) ... end` now generates warnings, I'm getting rid of these everywhere, in favor of the `where` notation
- [x] docstrings should be 1 newline away from what they refer to (i.e., no blank line in between, warning otherwise)
- [x] `findmax` does not work on generator expressions anymore

Package specific issues:
- [x] `dspev!` and `dspevV!` are broken

lostella:
Transition branch: https://github.com/kul-forbes/ProximalOperators.jl/tree/julia-0.7

nantonel:
This list is very useful! Thanks for sharing!

PR #59, opened 2018-08-18
Julia 1.0 update

lostella:
This PR addresses #47 and updates everything to Julia 1.0. Some notes:
* Before merging, a last v0.6-compatible version should be tagged;
* Some tests may fail due to `svds` until Arpack is updated, see https://github.com/JuliaLinearAlgebra/Arpack.jl/pull/32
* In particular, some hardcoded tests we had on `IndBallRank` don't work anymore, they are now commented and I'm about to look better into them;
* I've also added minor fixes to tests, most notably I **switched off all console outputs** to improve readability of error messages (except for the demo scripts, which are executed last and have very minor output which is supposed to stay there).

One of the very next things to do will be to sort out tests better: as they are now, it is rather difficult to include/exclude tests and to know exactly what caused what from the console output. What I imagine would work best is something along the lines of the newly added [test_linear.jl](https://github.com/kul-forbes/ProximalOperators.jl/blob/8950f7462fcb46230e9058ff5191ba4ebd415a7d/test/test_linear.jl), one test script per function type.

**Edit:** things may end up working on 0.7 only at first, since OSQP (or some of its dependencies) does not work on 1.0 just yet.

Things to do before merging:
- [x] Wait for Arpack.jl PR 32 **or**
- [x] Switch to TSVD.jl if that works better
- [x] Update README.md and docs
- [x] Tag the last 0.6-compatible version


nantonel (commented):
Maybe in the Readme we could also point to StructuredOptimization and the other packages!

lostella:
Switched from Arpack to TSVD for truncated singular value decomposition: TSVD implements the same algorithms as PROPACK, which are state of the art. In contrast, the truncated SVD based on ARPACK falls back into computing an equivalent symmetric eigenproblem, [which has known drawbacks](http://sun.stanford.edu/~rmunk/PROPACK/talk.rev3.pdf). From a quick benchmarking it seems to me that TSVD is both faster and more accurate.

Reintroduced all previously excluded tests for `IndBallRank`. Overall, tests pass on Julia 0.7, but fail on 1.0 due to OSQP.

@nantonel: I added a "Related packages" section in the readme.

nantonel:
@lostella that's great! I guess we just need wait for OSQP before merging?

lostella:
Maybe not: We can still allow failure for tests on 1.0 for the moment, and re-enable them when all dependencies are 1.0-ready.

The problem with waiting is that it could be a long wait: I think OSQP’s problems with 1.0 are due to its own dependencies, so who knows how long that could take. It’s maybe better to have at least a 0.7-ready version out.

src/calculus/moreauEnvelope.jl
```diff
@@ -33,17 +28,15 @@ is_quadratic(f::MoreauEnvelope) = is_generalized_quadratic(f.g)
...
+	buf = similar(x)
+	g_prox = prox!(buf, f.g, x, f.lambda)
+	return g_prox + 1/(2*f.lambda)*norm(buf .- x)^2
 end
 
 function gradient!(grad::AbstractArray, f::MoreauEnvelope, x::AbstractArray)
 	g_prox = prox!(grad, f.g, x, f.lambda)
-	grad .= (x - grad)/f.lambda
-	fx = g_prox + (f.lambda/2)*deepvecnorm(grad)^2
+	grad .= (x .- grad)/f.lambda
```
mfalt: I think that `./` might be more efficient here?
Example:
```julia
f1(x,y,z) = x .= (x .- y)/z
f2(x,y,z) = x .= (x .- y)/.z
x = randn(1000); y = randn(1000); z = 2.0
```
```julia
julia> @code_lowered f1(x,y,z)
CodeInfo(
1 1 ─ %1 = Base.Broadcast.materialize!                                                                            │
  │   %2 = Base.Broadcast.broadcasted                                                                             │
  │   %3 = Base.Broadcast.materialize                                                                             │
  │   %4 = Base.Broadcast.broadcasted                                                                             │
  │   %5 = (%4)(Main.:-, x, y)                                                                                    │
  │   %6 = (%3)(%5)                                                                                               │
  │   %7 = %6 / z                                                                                                 │
  │   %8 = (%2)(Base.identity, %7)                                                                                │
  │   %9 = (%1)(x, %8)                                                                                            │
  └──      return %9                                                                                              │
)
```
```julia
julia> @code_lowered f2(x,y,z)
CodeInfo(
1 1 ─ %1 = Base.Broadcast.materialize!                                                                            │
  │   %2 = Base.Broadcast.broadcasted                                                                             │
  │   %3 = Base.Broadcast.broadcasted                                                                             │
  │   %4 = (%3)(Main.:-, x, y)                                                                                    │
  │   %5 = (%2)(Main.:/, %4, z)                                                                                   │
  │   %6 = (%1)(x, %5)                                                                                            │
  └──      return %6                                                                                              │
)

```

where we see that `f1` has to materialize twice (tested on julia 0.7, with 2x difference in time).

lostella: Good catch! Fixing this

nantonel:
ok then let's merge it, I think it is practical also to update the other packages! Let's get 0.7 working everywhere!

lostella:
Agreed, let me do this later today, I just want to make a final round of checks

Issue #58, opened 2018-08-02
Missing functions and calculus rules


lostella:
This is a list of (some of the) currently missing functions and calculus rules for proximal mappings, with references. Please comment to add more to the list.

**Functions**
- [ ] Vapnik loss [Bauschke and Combettes, 2nd ed, Example 24.29][1]
- [x] Indicator of hyperslab, see [Bauschke and Combettes, 2nd ed, Example 29.21][1]
- [ ] Indicator of finite convex hull, see [Bauschke and Combettes, 2nd ed, Example 29.26][1]
- [x] Cubic Euclidean norm, see [Beck, Example 6.20][2]
- [ ] Negative Euclidean norm, see [Beck, Example 6.21][2]
- [ ] L1 norm over box, see [Beck, Example 6.23][2]
- [ ] Intersection of affine space and box [Beck, Theorem 6.27][2]
- [ ] Intersection of halfspace and box [Beck, Example 6.32][2]
- [ ] Intersection of L1 ball and box [Beck, Example 6.34][2]
- [ ] Sum of k largest values [Beck, Example 6.50][2] <= **THIS NEEDS FIXING**, see [sumLargest.jl](https://github.com/kul-forbes/ProximalOperators.jl/blob/master/src/functions/sumLargest.jl)
- [ ] Sum of k largest absolute values [Beck, Example 6.51][2]
- [ ] Phi and Bregman divergences, see [here](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=8194899)

**Calculus rules**
- [ ] Composition of even, smooth, convex function and distance from convex set, see [Bauschke and Combettes, 2nd ed, Proposition 24.27][1]
- [x] Prox of Moreau envelope, see [Beck, Theorem 6.63][2]
- [x] (#88) Epicomposition, see [epicompose.jl](https://github.com/kul-forbes/ProximalOperators.jl/blob/master/src/calculus/epicompose.jl)
- [ ] Spectral functions, see [Beck, Section 7][2]
    - [ ] Schatten p-norms
    - [ ] Ky Fan k-norms

**Notes**
* Some rules require functions to be even, so adding an `is_even` property may be convenient.

[1]: https://books.google.it/books/about/Convex_Analysis_and_Monotone_Operator_Th.html?id=PrM-DgAAQBAJ&redir_esc=y
[2]: https://sites.google.com/site/amirbeck314/books

mkrishan:
[...]

PR #62, opened 2018-11-11
adding IndHyperslab, fixing IndHalfspace and Linear

lostella:
This:
* adds `IndHyperslab`, partially addressing #58 (very partially)
* fixes `IndHalfspace` in that the improper case is detected, and `AbstractArray`s are allowed instead of just `AbstractVector`s
* fixes `Linear` in that only real coefficients are allowed
* makes parametric types a bit cleaner in all of the above


lostella:
Coverage drop due to https://github.com/JuliaCI/Coverage.jl/issues/187

Issue #64, opened 2018-11-30
Issue with views to cg!


baggepinnen:
There is an issue on [this line](https://github.com/kul-forbes/ProximalOperators.jl/blob/master/src/calculus/slicedSeparableSum.jl#L84) 
error message below
```
MethodError: Cannot `convert` an object of type Array{Float64,1} to an object of type SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}
Closest candidates are:
  convert(::Type{T<:AbstractArray}, !Matched::T<:AbstractArray) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T<:AbstractArray}, !Matched::Factorization) where T<:AbstractArray at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/factorization.jl:46
  convert(::Type{T}, !Matched::T) where T at essentials.jl:154
  ...
IterativeSolvers.CGStateVariables{Float64,SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}}(::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}) at cg.jl:111
cg!(::SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}, ::ProximalOperators.Shift{Array{Float64,2},Float64}, ::Array{Float64,1}) at cg.jl:210
prox!(::SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}, ::ProximalOperators.QuadraticIterative{Float64,Array{Float64,2},Array{Float64,1}}, ::SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}, ::Float64) at quadraticIterative.jl:29
prox!(::Array{Float64,1}, ::SlicedSeparableSum{Tuple{Array{ProximalOperators.QuadraticIterative{Float64,Array{Float64,2},Array{Float64,1}},1}},Array{Array{Tuple{UnitRange{Int64}},1},1},1}, ::Array{Float64,1}, ::Float64) at slicedSeparableSum.jl:84
[...]

PR #65, opened 2018-12-16
Bumping Julia requirement (bye bye 0.7)

lostella:
While [bumping the required version of IterativeSolvers](https://github.com/kul-forbes/ProximalOperators.jl/commit/4462acafe0311e4fe4a5cba764a10b5c8598aa13) to solve #64, I realized that they moved to requiring Julia 1.0. I guess it makes sense to start doing that move here too.

In addition:
* Added Project.toml (removed REQUIRE)
* Updated Travis and AppVeyor config files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,10 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
   - 1.0
   - nightly
 matrix:
   allow_failures:
-  - julia: 1.0
   - julia: nightly
 notifications:
   email: false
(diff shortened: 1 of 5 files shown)


nantonel:
All looks good! All the other packages of StructuredOptimization ecosystem are already in 1.0! 

By the way I see that now 'Project.toml' fully replaces 'REQUIRE'. 
I suppose we could do the same for the other packages...

nantonel (approved):

lostella:
Yes. However, as far as I understand Project.toml still needs to be edited manually in the “compat”, “extras”, and “target”.

Also, the rules for specifying compatible versions are slightly different for packages whose major version is 0. For example, here I will need to explicitly add compatibility with OSQP.jl v0.4 and v0.5 (adding them now)

lostella:
Epilogue: REQUIRE is still needed :-D

nantonel:
😂

PR #68, opened 2018-12-18
Fixing Documenter.jl setup

lostella:


lostella:
This seems not to fix it yet (see logs for the Linux/1.0 build on Travis). Also, when the setup will be ok we will still need to manually push the docs for v0.8 and v0.9 to the gh-pages branch

PR #69, opened 2019-01-05
Fixing types and indentation

lostella:
**Breaking:**
* Changed default constructor arguments for `IndBinary`: with zero arguments, it will still use `low=0.0` and `high=1.0`, but giving one argument only is not allowed anymore.

**Major:**
* Fixed typing issues in several functions (both in the signatures and in the body), and adding tests for some of them.

**Minor:**
* Indentation set to four spaces across most of `src/`, as suggested by
  1. https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md
  2. https://github.com/johnmyleswhite/Style.jl
* Typed 0, 1, and infinity literals are now uniformly denoted as `R(0)`, `R(1)`, and `R(Inf)` respectively, for `R <: Real`.
* Coveralls was removed from the Travis CI config, CodeCov is enough for the job: code coverage in Julia is currently flawed, so it doesn't make sense to report it to *two* services. This way we also lower the chance of getting unjustified red flags on PRs, like it has happened for several commits in this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,5 @@ matrix:
 notifications:
   email: false
 after_success:
-  - julia -e 'using Pkg; cd(Pkg.dir("ProximalOperators")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
   - julia -e 'using Pkg; cd(Pkg.dir("ProximalOperators")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
   - julia -e 'using Pkg; Pkg.add("Documenter"); cd(Pkg.dir("ProximalOperators")); include(joinpath("docs", "make.jl"))'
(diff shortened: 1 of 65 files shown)


mfalt:
I would prefer sticking to "4 spaces per indentation level, no tabs" as the contributing guidelines to base julia says. I assume most packages follows this?

lostella:
Good point, I was looking for something like it [here](https://docs.julialang.org/en/v1/manual/style-guide/index.html), but couldn't find much. Four spaces are also recommended here: https://github.com/johnmyleswhite/Style.jl. Updating!

lostella:
@mfalt I've updated the PR and the description. In particular, I think it is good to at least the default argument constructors as described above. I would actually be in favor of completely removing the defaults there. Thoughts?

I'm sorry for the numerous whitespace changes, you can turn them off in the diffs to see the changes better.

PR #76, opened 2019-03-27
WIP: First version of Polyhedral with QPDAS solver

mfalt:
QPDAS is still not registered so the tests will fail (but pass locally).

I ran a simple benchmark with vectors of length 1000, 50 equality, and 50 inequality constraints and no constraints on the vector itself. The benchmark results, running https://gist.github.com/mfalt/465969600b27a4ec285ab67c2530d374
are:
```julia
Setup OSQP
  0.015556 seconds (95 allocations: 2.365 MiB)
Setup QPDAS
  0.000734 seconds (128 allocations: 1.539 MiB)
First prox OSQP:
  0.698441 seconds (16 allocations: 25.938 KiB)
First prox QPDAS:
  0.000935 seconds (1.19 k allocations: 336.773 KiB)
100 prox! OSQP:
 68.826197 seconds (1.60 k allocations: 3.297 MiB)
100 prox! QPDAS:
  0.420764 seconds (591.76 k allocations: 38.232 MiB, 0.64% gc time)
```

Note that performance is worse when adding upper and lower constraints on the vector itself, I haven't tested how it compares to osqp in this case yet.

### DIFF
--- a/src/functions/indPolyhedralQPDAS.jl
+++ b/src/functions/indPolyhedralQPDAS.jl
@@ -1,0 +1,146 @@
+# IndPolyhedral: QPDAS implementation
+
+import QPDAS
+
+"""
+**Indicator of a polyhedral set**
+
+    IndPolyhedralQPDAS(A, C, b, d)
+
+```math
+S = \\{x : \\langle A, x \\rangle = b\\ ∧ \\langle C, x \\rangle \\le b\\}.
+```
+"""
+struct IndPolyhedralQPDAS{R<:Real, MT<:AbstractMatrix{R}, VT<:AbstractVector{R}, QP<:QPDAS.QuadraticProgram} <: IndPolyhedral
+    A::MT
+    b::VT
+    C::MT
+    d::VT
+    z::VT
+    qp::QP
+    first_prox::Ref{Bool}
+    function IndPolyhedralQPDAS{R}(A::MT, b::VT, C::MT, d::VT) where {R<:Real, MT<:AbstractMatrix{R}, VT<:AbstractVector{R}, QP<:QPDAS.QuadraticProgram}
+        @assert size(A,1) == size(b,1)
+        qp = QPDAS.QuadraticProgram(A, b, C, d, smartstart=false)
+        new{R, MT, VT, typeof(qp)}(A, b, C, d, zeros(R,size(A,2)), qp, Ref(true))
+    end
+end
+
+# properties
+
+is_prox_accurate(::IndPolyhedralQPDAS) = true
+
+# constructors
+
+function IndPolyhedralQPDAS(
+    l::AbstractVector{R}, A::AbstractMatrix{R}, u::AbstractVector{R}
+) where R
+    if !all(l .<= u)
+        error("function is improper (are some bounds inverted?)")
+    end
+    eqinds = (l .== u) .& .!isnothing.(l)
+    Aeq = A[eqinds,:]
+    beq = l[eqinds]
+
+    _islower(l::T) where T =
+        l != typemin(T) && !isnan(l) && !isnothing(l)
+    _isupper(u::T) where T =
+        u != typemax(T) && !isnan(u) && !isnothing(u)
+
+    lower = _islower.(l) .& (.!eqinds)
+    upper = _isupper.(u) .& (.!eqinds)
+
+    lower_only = lower .& (.! upper)
+    upper_only = upper .& (.! lower)
+    upper_and_lower = upper .& lower
+
+
+    Cieq = [-A[lower_only, :];
+             A[upper_only, :];
+            -A[upper_and_lower, :];
+             A[upper_and_lower, :] ]
+    dieq = [-l[lower_only];
+             u[upper_only];
+            -l[upper_and_lower];
+             u[upper_and_lower] ]
+
+    IndPolyhedralQPDAS{R}(Aeq, beq, Cieq, dieq)
+end
+
+IndPolyhedralQPDAS(
+    l::AbstractVector{R}, A::AbstractMatrix{R}, u::AbstractVector{R},
+    xmin::AbstractVector{R}, xmax::AbstractVector{R}
+) where R =
+    IndPolyhedralQPDAS([l; xmin], [A; I], [u; xmax])
+
+IndPolyhedralQPDAS(
+    l::AbstractVector{R}, A::AbstractMatrix{R}, args...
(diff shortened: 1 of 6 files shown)


src/functions/indPolyhedralQPDAS.jl
```diff
@@ -0,0 +1,146 @@
...
+             A[upper_only, :];
+            -A[upper_and_lower, :];
+             A[upper_and_lower, :] ]
+    dieq = [-l[lower_only];
+             u[upper_only];
+            -l[upper_and_lower];
+             u[upper_and_lower] ]
+
+    IndPolyhedralQPDAS{R}(Aeq, beq, Cieq, dieq)
+end
```
mfalt, line 68: This function is quite nasty, but I don't see another way to support the old syntax otherwise.

mfalt:
Didn't mean to push the benchmark, will remove those. 
This PR solves part of https://github.com/kul-forbes/ProximalOperators.jl/issues/52 (again)

src/functions/indPolyhedralQPDAS.jl
```diff
@@ -0,0 +1,146 @@
...
+
+import QPDAS
+
+"""
+**Indicator of a polyhedral set**
+
+    IndPolyhedralQPDAS(A, C, b, d)
+
+```math
+S = \\{x : \\langle A, x \\rangle = b\\ ∧ \\langle C, x \\rangle \\le b\\}.
```
lostella, line 11: I would write Ax and Cx without angular parentheses (we use those everywhere else for inner products)

src/functions/indPolyhedralQPDAS.jl
```diff
@@ -0,0 +1,146 @@
...
+    IndPolyhedralQPDAS(
+        R(-Inf).*ones(R, size(A, 1)), A, u, args...
+    )
+
+# function evaluation
+
+function (f::IndPolyhedralQPDAS{R})(x::AbstractVector{R}) where R
+    Ax = f.A * x
+    Cx = f.C * x
+    return all(Ax .<= f.b .& Cx .<= f.d) ? R(0) : Inf
```
lostella, line 95: Is it `Ax .== f.b` here? Or rather `isapprox.(Ax, f.b)`.

If this was a bug and no test failed, maybe a test is missing

src/functions/indPolyhedralQPDAS.jl
lostella: It’s okay, I don’t see many other way of doing it.

I’m open to changing the overall syntax for the construction of `IndPolyhedral` objects. But maybe we can open a dedicated issue for that and focus here in meeting the current construction signature.

lostella (changes requested):
Thank you! Benchmark results are impressive, good job with the solver! 

Could you also update `IndPolyhedral` as part of this, and add the option for using this variant? Also, the docstring for `IndPolyhedral` could be updated with the solver argument and a brief description of the available options for it. I believe I didn’t do this before since OSQP was the only option there.

QPDAS is already on its way to be registered?

src/functions/indPolyhedralQPDAS.jl
mfalt: You're right, I just copied or from somewhere and forgot to change that.

src/functions/indPolyhedralQPDAS.jl
mfalt: Bug, will fix and add test.

mfalt:
> 
> Could you also update `IndPolyhedral` as part of this, and add the option for using this variant? 

This should already be in the PR, or did you mean something else?
>
> Also, the docstring for `IndPolyhedral` could be updated with the solver argument and a brief description of the available options for it. I believe I didn’t do this before since OSQP was the only option there.

Will do
> 
> QPDAS is already on its way to be registered?

I opened it to the public yesterday, wanted to test it against osqp here first. Will probably request it to be registered this weekend.

I will also post some more benchmarks, it seems like it is consistently 2x to 3000x faster than osqp for a wide variety of projection problems. So we should probably considering to make it the new dafault, maybe through some deprecation? (Trying to not be too partial here)

lostella:
Yes sorry, it’s already in this PR, I was probably checking in master by mistake.

I’m open to changing the default, let’s maybe look into a more extensive comparison for this type of projections, including ill-conditioned problems where some solvers may have a hard time.

By the way, are you aware of any other open source QP solvers with Julia bindings (or pure Julia ones, even better)? Because when I was looking into this, I could only find OSQP, which was also very young back then.

mfalt:
> I’m open to changing the default, let’s maybe look into a more extensive comparison for this type of projections, including ill-conditioned problems where some solvers may have a hard time.

I completely agree, I'm currently running tests for large problems (its currently at 10_000 variables, 2000 equality, 2000inequality, osqp takes 2hours, qpdas 20min (doing 10 consecutive proxes on similar points). In this case I think the dual is quite ill-conditioned, do you have some good tips for generating ill-conditioned problems?

> By the way, are you aware of any other open source QP solvers with Julia bindings (or pure Julia ones, even better)? Because when I was looking into this, I could only find OSQP, which was also very young back then.

Not really, at least not active-set solvers. qpOASES has similar performance as QPDAS, but it is in C++, and I was unable to write a working wrapper since it keeps variables in memory internally.

mfalt:
The benchmarks for random problems with bounds either as `lb <= Ax <= ub`, (denoted u+l) and `Ax <= ub` (u) are as follows on my computer, for n=100 to 10000 variables:
osqp_qpdas_benchmark_random

where the ratios of the timings:
osqp_qpdas_benchmark_ratio

Note that QPDAS is faster in almost all cases, except when the number of variables is 5000 or 10000, and there are 2500 and 5000 upper and lower constraints respectively. However, osqp takes 2.2 and 11 hours to do init + 11proxes, and qpdas 0.5 and 8 hours in these cases, so I am not sure it is relevant for this package. 

I will try to generate more tests for ill-conditioned problems when I have time.

hakkelt:
Hey @mfalt, do you still plan to finalize this PR?

mfalt:
@hakkelt I wasn't planning on it. I would say that the main problem isn't finishing the implementation, that should be easy, but actually verifying if the solver is good enough on a wide range of problems. I'm happy to finish the PR if lostella believes the solver to be good enough, but I don't really have the time to evaluate that myself.

hakkelt:
Ok, thanks for the answer.

PR #77, opened 2019-03-31
Fixes for 1.2

lostella:
This PR fixes a couple of issues that currently fail tests on the soon-to-be Julia 1.2

### DIFF
--- a/src/functions/leastSquaresIterative.jl
+++ b/src/functions/leastSquaresIterative.jl
@@ -48,6 +48,7 @@ function prox!(y::AbstractVector{D}, f::LeastSquaresIterative{R, RC, M, V}, x::A
     f.q .= f.Atb .+ x./lamgam
     # two cases: (1) tall A, (2) fat A
     if f.shape == :Tall
+        y .= x
         op = Shift(f.S, R(1)/lamgam)
         IterativeSolvers.cg!(y, op, f.q)
     else # f.shape == :Fat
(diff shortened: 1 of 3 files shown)


src/functions/quadraticIterative.jl
```diff
@@ -24,6 +24,7 @@ function (f::QuadraticIterative{R, M, V})(x::AbstractArray{R}) where {R, M, V}
 end
 
 function prox!(y::AbstractArray{R}, f::QuadraticIterative{R, M, V}, x::AbstractArray{R}, gamma::R=R(1)) where {R, M, V}
+    y .= x
```
lostella, line 27: This is required since if `y = similar(x)` then `y` could contain `NaN`, and apparently this occurs in tests when running on Julia 1.2: it makes sense that his happens I guess, since `NaN` is a valid floating point value. 

Note that this change also has the effect of initializing `cg!` (see below) with `x` as starting point, which makes perfect sense to me.

src/functions/leastSquaresIterative.jl
```diff
@@ -48,6 +48,7 @@ function prox!(y::AbstractVector{D}, f::LeastSquaresIterative{R, RC, M, V}, x::A
     f.q .= f.Atb .+ x./lamgam
     # two cases: (1) tall A, (2) fat A
     if f.shape == :Tall
+        y .= x
```
lostella, line 51: See comment on quadraticIterative.jl below, same story goes here.

test/test_graph.jl
```diff
@@ -12,17 +12,17 @@ sm = n + div(n, 10)
...
 
     ## TODO: remove when IndAffine will be revised
-    if T <: Complex{Float64}
-        return 0.0
+    if T <: Complex
+        return
     end
 
-    B = ifelse(issparse(A), [A -SparseMatrixCSC(I, m, m)],  [A -Matrix(I, m, m)])
+    B = ifelse(issparse(A), [A -SparseMatrixCSC{T}(I, m, m)],  [A -Matrix{T}(I, m, m)])
```
lostella, line 25: This line was causing issues in the tests.

PR #83, opened 2019-09-24
Added Transpose and Adjoint functions for LeastSquaresDirect

mfalt:
`LeastSquares(A', b, lambda)` threw an error since `warn` is not defined.

Updated to `@warn` macro to fix bug, and added case to catch Adjoint and Transposes by dong a copy.

Copy was needed since the type is defined to have:
`A::M` and `S::M` where `S = A'*A`, which does not hold for transposes/adjoints.

### DIFF
--- a/src/ProximalOperators.jl
+++ b/src/ProximalOperators.jl
@@ -11,6 +11,7 @@ const ArrayOrTuple{R} = Union{
     AbstractArray{C, N} where {C <: RealOrComplex{R}, N},
     TupleOfArrays{R}
 }
+const TransposeOrAdjoint{M} = Union{Transpose{C,M} where C, Adjoint{C,M} where C}
 
 export ProximableFunction
 export prox, prox!, gradient, gradient!
--- a/src/functions/leastSquaresDirect.jl
+++ b/src/functions/leastSquaresDirect.jl
@@ -108,7 +108,7 @@ function solve_step!(y::AbstractVector{C}, f::LeastSquaresDirect{R, C, M, V, F},
     end
 end
 
-function solve_step!(y::AbstractVector{C}, f::LeastSquaresDirect{R, C, M, V, F}, x::AbstractVector{C}, gamma::R) where {R, C, M, V, F <: SuiteSparse.CHOLMOD.Factor{C}}
+function solve_step!(y::AbstractVector{C}, f::LeastSquaresDirect{R, C, M, V, F}, x::AbstractVector{C}, gamma::R) where {R, C, M, V, F}
     lamgam = f.lambda*gamma
     f.q .= f.Atb .+ x./lamgam
     # two cases: (1) tall A, (2) fat A
--- a/src/functions/leastSquaresDirect.jl
+++ b/src/functions/leastSquaresDirect.jl
@@ -76,7 +76,7 @@ function prox!(y::AbstractVector{C}, f::LeastSquaresDirect{R, C, M, V, F}, x::Ab
     return (f.lambda/2)*norm(f.res, 2)^2
 end
 
-function factor_step!(f::LeastSquaresDirect{R, C, M, V, F}, gamma::R) where {R, C, M <: DenseMatrix, V, F}
+function factor_step!(f::LeastSquaresDirect{R, C, M, V, F}, gamma::R) where {R, C, M, V, F}
     lamgam = f.lambda*gamma
     f.fact = cholesky(f.S + I/lamgam)
     f.gamma = gamma
--- a/src/functions/leastSquaresDirect.jl
+++ b/src/functions/leastSquaresDirect.jl
@@ -45,8 +45,17 @@ function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{
     LeastSquaresDirect{R, C, M, V, SuiteSparse.CHOLMOD.Factor{C}}(A, b, lambda)
 end
 
+# Adjoint/Transpose versions
+function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{R, Complex{R}}, M <: TransposeOrAdjoint{<:DenseMatrix{C}}, V <: AbstractVector{C}}
+    LeastSquaresDirect(copy(A), b, lambda)
+end
+
+function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{R, Complex{R}}, I <: Integer, M <: TransposeOrAdjoint{<:SparseMatrixCSC{C, I}}, V <: AbstractVector{C}}
+    LeastSquaresDirect(copy(A), b, lambda)
+end
+
 function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{R, Complex{R}}, M <: AbstractMatrix{C}, V <: AbstractVector{C}}
-    warn("Could not infer type of Factorization for $M in LeastSquaresDirect, this type will be type-unstable")
+    @warn "Could not infer type of Factorization for $M in LeastSquaresDirect, this type will be type-unstable"
     LeastSquaresDirect{R, C, M, V, Factorization}(A, b, lambda)
 end
 
--- a/test/test_leastSquares.jl
+++ b/test/test_leastSquares.jl
@@ -261,3 +261,70 @@ grad_gx, gx = gradient(g, x)
 call_test(g, x)
 prox_test(g, x)
 prox_test(g, x, 2.1)
+
+# Tall full adjoint matrix
+
+m, n = 30, 10
+A = randn(n, m)
+b = randn(m)
+x = randn(n)
+
+f = LeastSquares(A', b)
+@test ProximalOperators.is_smooth(f) == true
+@test ProximalOperators.is_quadratic(f) == true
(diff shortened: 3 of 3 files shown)


src/ProximalOperators.jl
```diff
@@ -11,6 +11,7 @@ const ArrayOrTuple{R} = Union{
     AbstractArray{C, N} where {C <: RealOrComplex{R}, N},
     TupleOfArrays{R}
 }
+const TransposeOrAdjoint{M} = Union{Transpose{C,M} where C, Adjoint{C,M} where C}
```
mfalt, line 14: I see no reason to restrict `C` in this case since it is already restricted (defined) by `eltype(M)`, but I'm open to suggestions. The only case I can see where something might possibly go wrong is if `eltype(A) != eltype(A')`, but in that case I think we are in big trouble already by assuming things on `A'A`.

src/functions/leastSquaresDirect.jl
```diff
@@ -45,8 +45,17 @@ function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{
     LeastSquaresDirect{R, C, M, V, SuiteSparse.CHOLMOD.Factor{C}}(A, b, lambda)
 end
 
+# Adjoint/Transpose versions
+function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{R, Complex{R}}, M <: TransposeOrAdjoint{<:DenseMatrix{C}}, V <: AbstractVector{C}}
+    LeastSquaresDirect(copy(A), b, lambda)
+end
+
+function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{R, Complex{R}}, I <: Integer, M <: TransposeOrAdjoint{<:SparseMatrixCSC{C, I}}, V <: AbstractVector{C}}
+    LeastSquaresDirect(copy(A), b, lambda)
```
lostella, line 54: Why the need to `copy` here (similarly in line 50)? If I’m not wrong, `adjoint(adjoint(X)) === X` for any `X` (same for `transpose`), see

https://github.com/JuliaLang/julia/blob/38b38cfdd5b1a5bdf1a613a6180d2aaadf06a632/stdlib/LinearAlgebra/src/adjtrans.jl#L152-L153

so maybe we can use `adjoint(A)` here and avoid further allocations? Of course, that would require that the `Transpose` and `Adjoint` cases are coded separately.

lostella (commented):
Looks good, thanks for catching this. I only have one concern about allocations, see inlined comments

src/ProximalOperators.jl
lostella: Yeah I think this is fine

src/functions/leastSquaresDirect.jl
lostella: Actually, one could use `A.parent` and catch both `Transpose` and `Adjoint` with one implementation, but that would rely on the internals of some structure owned by others, which seems weird to do. I’m not sure 🤔

mfalt: The function is catching the case
`LeastSquares(X', b, lambda)` or `LeastSquares(transpose(X), b, lambda)`
i.e where we want to work with the adjoint of some allocated matrix.
The way I see it is that we have three options if we dont want to  `copy(A)` i.e. `copy(X')`:

* We save `A=X`, and have a flag for `adjoint/transpose/identity` and then have separate implementations based on that flag for `prox!`, `factor_step!`, `solve_step!`, `gradient!`.
* We save A=X' in the struct, which means that `A::Adjoint{Float64,Matrix{Float64}}`, for example. But then we no longer have that `S = A'*A`, is of type `M` so we need a new type parameter, say M2. We want to be able to infer the factorization type `F` in this case, which for the ones we already inffer I guess is `Cholesky{C,M2}`, and `SuiteSparse.CHOLMOD.Factor{C}` anyway.
* We only fix `warn()` -> `@warn`. which will result in a LeastSquares with `F=Factorization`. But it doesn't even look like `prox!` will work in this case, I cant see a version of `solve_step!` that would catch it. It would probably be possible to write a type-unstable constructor instead that creates a LeastSquares with the proper `F`, but again, this requires a bit of re-coding.

The second one I think would be all right. But in either case, the constructor allocates both `S = A'*A` or `S = A*A'` as well as a factorization and a matrix in the first `prox!`: `cholesky(f.S + I/lamgam)`.
To be fair, if `A` is really tall or fat, these allocations will "only" be `min(m,n)^2` compared to a `copy(A)` which is `m*n`, but I'm not sure that one could argue that `copy(A)` is really that much worse than what we are already doing. The best argument I can find is that if the user only wants to compute gradients, then the constructor could have been allocation free (except for a vector), but that would require a larger rewrite for the case we already support.

Edit: Changed some A to X

mfalt:
As a slightly separete issue, the function
https://github.com/kul-forbes/ProximalOperators.jl/blob/fc4035de3c339d19428e14203af696ee20a777ef/src/functions/leastSquaresDirect.jl#L111
doesn't use that `F <: SuiteSparse.CHOLMOD.Factor{C}`, so it could probably be used as a fallback for the general case `F <: Factorization`

lostella (approved):

src/functions/leastSquaresDirect.jl
lostella: Disregard my comment, I was wrong. It’s fine to use `copy`, since that’s definitely correct.

lostella:
@mfalt feel free to remove that restriction on `F` in the `solve_step!` before merging! And thanks for the PR!

lostella:
I’ll cut a new release as soon as this is merged, unless you believe that there are other urgent fixes to make

mfalt:
I was just about to but remembered that there was something weird going on before if lambda is an integer, throwing "R" not defined. Can you figure out what is going on here? It is a problem since before the PR:
```julia
function LeastSquaresDirect(A::M, b::V, lambda::R) where {R <: Real, C <: Union{R, Complex{R}}, M <: DenseMatrix{C}, V <: AbstractVector{C}}
    println("M: $M")
    println("V: $V")
    println("C: $C")
    println("R: $R")
    LeastSquaresDirect{R, C, M, V, Cholesky{C, M}}(A, b, lambda)
end

f =  LeastSquares(X, b, 1)

M: Array{Float64,2}
V: Array{Float64,1}
C: Float64
ERROR: UndefVarError: R not defined
Stacktrace:
 [1] ProximalOperators.LeastSquaresDirect(::Array{Float64,2}, ::Array{Float64,1}, ::Int64) at /local/home/mattiasf/.julia/dev/ProximalOperators/src/functions/leastSquaresDirect.jl:44
 [2] #LeastSquares#4(::Bool, ::Type{LeastSquares}, ::Array{Float64,2}, ::Array{Float64,1}, ::Int64) at /local/home/mattiasf/.julia/dev/ProximalOperators/src/functions/leastSquares.jl:31
 [3] LeastSquares(::Array{Float64,2}, ::Array{Float64,1}, ::Int64) at /local/home/mattiasf/.julia/dev/ProximalOperators/src/functions/leastSquares.jl:30
 [4] top-level scope at none:0
```

I don't understand how it can both catch the call and throw that error.
julia> versioninfo()
Julia Version 1.2.0-rc2.27
I will try again with the release.

lostella:
> I don't understand how it can both catch the call and throw that error.
> julia> versioninfo()
> Julia Version 1.2.0-rc2.27
> I will try again with the release.

The same problem occurs on Julia 1.2.0. Do you think we can anyway merge the PR and address this in a separate story?

mfalt:
Sure, let me fix the comment you made and then we can pull and tag. I have about 50 students who are new to Julia and running into these weird error messages...

mfalt:
> > I don't understand how it can both catch the call and throw that error.
> > julia> versioninfo()
> > Julia Version 1.2.0-rc2.27
> > I will try again with the release.
> 
> The same problem occurs on Julia 1.2.0. Do you think we can anyway merge the PR and address this in a separate story?

Created a Julia issue, since this seems to be a julia problem. https://github.com/JuliaLang/julia/issues/33381
However, we probably want to allow `lambda=1` as input somehow, so maybe we should have a separate issue for this?

lostella:
> However, we probably want to allow `lambda=1` as input somehow, so maybe we should have a separate issue for this?

Yes, in the meantime a workaround is to just use `lambda=1.0` I guess

src/functions/leastSquaresDirect.jl
```diff
@@ -67,7 +76,7 @@ function prox!(y::AbstractVector{C}, f::LeastSquaresDirect{R, C, M, V, F}, x::Ab
     return (f.lambda/2)*norm(f.res, 2)^2
 end
 
-function factor_step!(f::LeastSquaresDirect{R, C, M, V, F}, gamma::R) where {R, C, M <: DenseMatrix, V, F}
+function factor_step!(f::LeastSquaresDirect{R, C, M, V, F}, gamma::R) where {R, C, M, V, F}
```
lostella, line 79: I thought you meant removing the restriction on `F` in line 111 below

mfalt: You are right, I missed that. But I think both are decent fallbacks, and they are both needed if we should have any chance of solving for general types. Added the change in line 111 also

lostella (approved):

Issue #74, opened 2019-03-12
Return types should agree with inputs?


lostella:
We currently have a strange contract regarding return types when calling functions, `gradient`, and `prox`, which should probably simplified.

`versioninfo()`:

```
Julia Version 1.1.0
Commit 80516ca202 (2019-01-21 21:24 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-7660U CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

If one does

```julia
using ProximalOperators
f = SqrNormL2(Float64(1.0))
x = randn(Float32, 10)
```

then

```julia
julia> typeof(f(x))
Float64

julia> g, v = gradient(f, x); typeof(v)
Float64

julia> p, v = prox(f, x); typeof(v)
Float64
```

The reason is that the default coefficient in `SqrNormL2` is `Float64(1.0)`. Things are potentially more explicit and predictable if function values are forced to be the same as `real(eltype(x))` (where `real` secures the case where `eltype(x) <: Complex`).

It would be nice to sort this out and make sure tests cover this. And also document this!

Issue #16, opened 2017-02-14
An immutable type ProximalPoint


lostella:
This is just a design idea, but I kinda like it.

From the beginning of ProximalOperators, we had `prox!` and `prox` returning both the proximal point `y` *and* the value of `f` at `y`. This has several advantages: essentially, in many cases evaluating the prox of a function entails almost-computing the function value at the resulting point. Such value may be used, for example, in a stopping criterion, or may be of any other use for an algorithm. A notable example for this is the nuclear norm: once you have soft-thresholded the singular values, you better store their sum somewhere: otherwise evaluating `f` at the resulting point will be another SVD - quite expensive.

Actually, right now `prox!` only returns the function value, since the proximal point is written to an array provided by the user.

On the other hand, I can see the advantage of having the proximal point as the only returned argument. It's a bit more like it is in paper: you ask for the proximal point, you get the proximal point.

One way I can imagine to obviate this problem is to have an `Array`-like type, encapsulating the function value along with the proximal point. It should be `immutable` by law (point & function value shall be together until the last bit of their lives, in the name of provable correctness).

```julia
immutable ProximalPoint <: AbstractArray
  f::ProximableFunction
  point::AbstractArray
  value::Real
end
```

An object of this type shall be created upon prox evaluation. Then the `call` method would be:

```julia
function (f::ProximableFunction)(x::ProximalPoint)
  if object_id(f) == object_id(x.f)
    return x.value
  else
    return f(x.point) # this is already defined for all functions
  end
end
```

This way one can decide to do:

```julia
y = prox(f, x, gamma)
v = f(y) # this doesn't compute anything
```

Would this complicate things? That is, would this require us to (trivially) redefine several other things for the new `ProximalPoint` type? Things that, when called on `x::ProximalPoint`, should be re-called with argument `x.point` instead. It would be nice if it were possible to declare: whatever acts on `AbstractArray` acts on `ProximalPoint` by looking at its `point` field (unless otherwise specified: thank you, multiple dispatch).

mfalt:
I think this is a very interesting idea, but I think we have to think it through really carefully. I think there are a couple of trade-offs regarding how much data should be saved to be efficient on multiple calls versus being efficient on a single call.
Do you think this should be the output even for `prox!`? This could potentially lead to a lot of extra allocations, for example in
```julia
for i = 1:100000
   prox!(y,f,x)
   x = 0.5*x .+ 0.5*y
end
```
[...]

PR #89, opened 2019-12-27
Making function value types the same as input eltype

lostella:
~This PR aims at addressing #74, as well as factoring a few tests a bit better as a side effect.~

Essentially, this enforces the following

```julia
v1 = f(x)
_, v2 = prox(f, x)
_, v3 = gradient(f, x)
typeof(v1) == typeof(v2) == typeof(v3) == real(eltype(x))
```

for any `f::ProximableFunction`, *whenever `f` is constructed with default arguments, or with parameters based off real(eltype(x))*.

**Description of the changes**

The above is achieved by:
* lowering the default function parameters to integer
* asserting the above condition, e.g. in the `prox_test` utility function *(this was there already, but tests were covering `Float64` only AFAIU)*
* enabling some test cases with multiple underlying real types, e.g. using `Float32` and `Float64`, or the corresponding complex types *(this required changes to test_calls.jl, and I took the chance to also reorganize that script quite a bit)*

**Remarks**

~In principle, one could *additionally* require the `real(eltype(x))` to be the same as the real type underlying the parameters of `f` (such as coefficients), and avoid the need of doing type conversion at all. However, this creates issues when the parameters have defaults: for example, the coefficient `lambda` in `f = NormL1()` defaults to `1.0::Float64` (on 64-bit machines, I guess), so by default `f` would only be applicable to `AbstractArray{Float64}` or `AbstractArray{Complex{Float64}}`, and one would need to explicitly construct `f = NormL1(Float32(1))` to be able to use it with `x::Array{Float32}`.~

~Note also that not all test cases have been enabled with multiple real types just yet, I'm opening this PR nevertheless to get early feedback.~

@mfalt I remember we discussed this in the past: I'd appreciate inputs from you on this


Project.toml
```diff
@@ -11,9 +11,9 @@ SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 TSVD = "9449cd9e-2762-5aa3-a617-5413e99d722e"
 
 [compat]
-IterativeSolvers = ">= 0.8.0"
-OSQP = ">= 0.3.0"
-TSVD = ">= 0.3.0"
+IterativeSolvers = "0.8"
+OSQP = "0.3, 0.4, 0.5, 0.6"
+TSVD = "0.3, 0.4"
```
lostella: These are needed since dependencies versions must be upper bounded in order for packages to be automatically registered. See some context here: https://discourse.julialang.org/t/please-be-mindful-of-version-bounds-and-semantic-versioning-when-tagging-your-packages/30708

mfalt:
I am not completely convinced by #74 but this might be a good idea. I will have to look through it a bit more carefully beginning next week.

If possible, it would be really nice to test with weird things like dual numbers. I'm afraid this could break some automatic differentiability (without looking at the code).

Project.toml
mfalt: We should probably install the compat helper bot. It will create a PR whenever a dependency releases a new major version. (And we should then freely release patch versions of our package). It is very annoying for a user if it installs OSQP first, gets the newest version, and then is unable to install ProximalOperators.

lostella:
> If possible, it would be really nice to test with weird things like dual numbers. I'm afraid this could break some automatic differentiability (without looking at the code).

I agree that we could test those situations. But this PR shouldn’t (and hopefully currently doesn’t) affect much the manipulation of `x`. For example, no type conversion occurs on `x` or other arrays derived from `x`, but only on function parameters as soon as they enter the computational graph.

lostella:
@mfalt do you have any further concern with this? Otherwise I would like to get this merged soon. The idea in #74 makes sense to me: for any function `f`, if `x` is a `FloatXYZ`-based array then `f(x)` will be of type `FloatXYZ`. See my previous comment about autodiff:

> this PR shouldn’t (and hopefully currently doesn’t) affect much the manipulation of x. For example, no type conversion occurs on x or other arrays derived from x, but only on function parameters as soon as they enter the computational graph

mfalt:
Sorry for not looking at this earlier. I think it looks very reasonable, especially since the conversions are only applied to the parameters. I think most of this problem could be solved by making sure that all defaults are low enough, for example integers, but it would leave no guarantees like this does. The parameters seems to be mostly restricted to Reals anyway, so there might not be any (interesting) existing cases where the conversion fails (for example the parameter being a dual number).

The only case I can come up with that breaks is if the parameter is a float and the user tries to call the function or prox with integer values, but that's not a big deal.

My general feeling is that if a user creates a function with Float64 parameters, then the user is also happy with receiving Float64 outputs. But I have no strong opinions, so feel free to merge. I definetely don't think it's bad, maybe slightly unnecessary. But I also don't see that the PR would cause any problems.

mfalt (approved):

lostella:
> (for example the parameter being a dual number).

Ok, I thought you referred to the input, not parameters, when mentioning dual numbers. Yes, they are mostly restricted to Real for now: *that* might be something that we want to remove, in which case we can rethink this. The idea of having low enough default parameters may be ideal then.

> The only case I can come up with that breaks is if the parameter is a float and the user tries to call the function or prox with integer values, but that's not a big deal.

For that, would it make sense to restrict `x` to `AbstractFloat`-based arrays?

mfalt:
> For that, would it make sense to restrict `x` to `AbstractFloat`-based arrays?

I think it's better to keep the restriction on x as it is in this PR. I would rather get accuracy errors in some special cases than being unable to call the function for some types where it would actually work.

Edit: e.g FixedPoint or Rational

lostella:
@mfalt you're right. In any case, I've given your previous comments some thought (thanks for those) and I'm convinced that all this casting of function parameters may not be the best solution. I'm working instead towards lowering the type of their defaults, as you suggested, and it appears much cleaner to me. With that, the convention on input vs function type would be

> `typeof(f(x)) == real(eltype(x))` whenever `f` is instantiated with default options, or with parameters based off `real(eltype(x))`

Similarly for `prox`. This is a weaker condition than #74, does not involve any type conversion, and is mainly about having test cases that cover the above condition. Also there were still many cases uncovered in tests here.

My available time is what it is, but I'll update this PR as soon as possible.

mfalt:
I think that sounds like a good compromise, hopefully it's not too tricky to implement. Unfortunately I don't currently have any spare time to spend on this project.

lostella:
Done: with the latest update, now the changes to the package should really be minor (albeit the diff is long). Most of the elaborate changes are really in the tests.

Kindly asking @nantonel to also take a look :-) disable white-space changes for easier review!

mfalt:
I skimmed through the changes and it looks really good. We should just make sure to tag it 0.13 some of the changed type parameters might be breaking for someone, although it seems unlikely.

Is there a roadmap for bumping the version to 1.0? It feels the package is very stable by now.
Edit: In fact, it seems like the only open issue that would be breaking enough for is to wait with tagging 1.0 is https://github.com/kul-forbes/ProximalOperators.jl/issues/16 . I wasn't convinced by it at the time, but it was a while ago since I considered it.

src/calculus/distL2.jl
```diff
@@ -30,34 +30,34 @@ end
...
+    d = normdiff(x, y)
+    gamlam = gamma * f.lambda
     if gamlam < d
         gamlamd = gamlam/d
-        y .= (1-gamlamd).*x .+ gamlamd.*y
-        return f.lambda*(d-gamlam)
+        y .= (1 - gamlamd) .*x .+ gamlamd .* y
+        return f.lambda * (d - gamlam)
     end
     return R(0)
```
nantonel, line 49: `zero(R)`?

src/functions/huberLoss.jl
```diff
@@ -33,41 +33,41 @@ end
...
-        return (f.mu/2)*normx^2
+        return f.mu / R(2) * normx^2
     else
-        return f.rho*f.mu*(normx-f.rho/2)
+        return f.rho * f.mu * (normx - f.rho / R(2))
     end
 end
 
-function gradient!(y::AbstractArray{T}, f::HuberLoss, x::AbstractArray{T}) where T <: Union{Real, Complex}
+function gradient!(y::AbstractArray{T}, f::HuberLoss, x::AbstractArray{T}) where {R <: Real, T <: Union{R, Complex{R}}}
```
nantonel, line 47: `T<:RealOrComplex{R}`

src/functions/huberLoss.jl
```diff
@@ -33,41 +33,41 @@ end
...
-        y .= (f.mu*f.rho)/normx .* x
-        v = f.rho*f.mu*(normx-f.rho/2)
+        y .= (f.mu * f.rho) / normx .* x
+        v = f.rho * f.mu * (normx - f.rho / R(2))
     end
     return v
 end
 
-function prox!(y::AbstractArray{T}, f::HuberLoss, x::AbstractArray{T}, gamma::Real=1.0) where T <: Union{Real, Complex}
+function prox!(y::AbstractArray{T}, f::HuberLoss, x::AbstractArray{T}, gamma::R=R(1)) where {R <: Real, T <: Union{R, Complex{R}}}
```
nantonel, line 59: `T<:RealOrComplex{R}`

src/functions/huberLoss.jl
```diff
@@ -76,11 +76,11 @@ fun_dom(f::HuberLoss) = "AbstractArray{Real}, AbstractArray{Complex}"
 fun_expr(f::HuberLoss) = "x ↦ (μ/2)||x||² if ||x||⩽ρ, μρ(||x||-ρ/2) otherwise"
 fun_params(f::HuberLoss) = string("ρ = $(f.rho), μ = $(f.mu)")
 
-function prox_naive(f::HuberLoss, x::AbstractArray{T}, gamma::Real=1.0) where T <: Union{Real, Complex}
-    y = (1-min(f.mu*gamma/(1+f.mu*gamma), f.mu*gamma*f.rho/(norm(x))))*x
+function prox_naive(f::HuberLoss, x::AbstractArray{T}, gamma::R=R(1)) where {R <: Real, T <: Union{R, Complex{R}}}
```
nantonel, line 79: `T<:RealOrComplex{R}`

src/functions/indExp.jl
```diff
@@ -28,27 +28,27 @@ Returns the indicator function of the dual exponential cone, that is
...
 EXP_POLAR_CALL_TOL = 1e-3
 EXP_PROJ_TOL = 1e-15
 EXP_PROJ_MAXIT = 100
 
 function (f::IndExpPrimal)(x::AbstractArray{R,1}) where R <: Real
-    if (x[2] > 0.0 && x[2]*exp(x[1]/x[2]) <= x[3]+EXP_PRIMAL_CALL_TOL) ||
+    if (x[2] > 0 && x[2]*exp(x[1]/x[2]) <= x[3]+EXP_PRIMAL_CALL_TOL) ||
         (x[1] <= EXP_PRIMAL_CALL_TOL && abs(x[2]) <= EXP_PRIMAL_CALL_TOL && x[3] >= -EXP_PRIMAL_CALL_TOL)
-        return 0.0
+        return R(0)
```
nantonel, line 41: zero(R)

src/functions/indExp.jl
```diff
@@ -28,27 +28,27 @@ Returns the indicator function of the dual exponential cone, that is
...
-    return +Inf
+    return R(Inf)
 end
 
 function (f::Conjugate{IndExpPrimal})(x::AbstractArray{R,1}) where R <: Real
-    if (x[1] > 0.0 && x[1]*exp(x[2]/x[1]) <= -exp(1)*x[3]+EXP_POLAR_CALL_TOL) ||
+    if (x[1] > 0 && x[1]*exp(x[2]/x[1]) <= -exp(1)*x[3]+EXP_POLAR_CALL_TOL) ||
         (abs(x[1]) <= EXP_POLAR_CALL_TOL && x[2] <= EXP_POLAR_CALL_TOL && x[3] <= EXP_POLAR_CALL_TOL)
-        return 0.0
+        return R(0)
```
nantonel, line 49: zero(R)

src/functions/indPoint.jl
```diff
@@ -28,14 +28,14 @@ is_affine(f::IndPoint) = true
...
-function (f::IndPoint{R, C})(x::AbstractArray{C}) where {R, C}
+function (f::IndPoint)(x::AbstractArray{C}) where {R <: Real, C <: Union{R, Complex{R}}}
     if all(x .≈ f.p)
         return R(0)
     end
     return R(Inf)
 end
 
-function prox!(y::AbstractArray{C}, f::IndPoint{R, C}, x::AbstractArray{C}, gamma=R(1)) where {R, C}
+function prox!(y::AbstractArray{C}, f::IndPoint, x::AbstractArray{C}, gamma=R(1)) where {R <: Real, C <: Union{R, Complex{R}}}
```
nantonel, line 38: `C<:RealOrComplex{R}`

src/functions/indPoint.jl
```diff
@@ -46,7 +46,7 @@ fun_expr(f::IndPoint) = "x ↦ 0 if x = p, +∞ otherwise"
 fun_params(f::IndPoint) =
     string( "p = ", typeof(f.p) <: AbstractArray ? string(typeof(f.p), " of size ", size(f.p)) : f.p, ", ")
 
-function prox_naive(f::IndPoint{R, C}, x::AbstractArray{C}, gamma=R(1)) where {R, C}
+function prox_naive(f::IndPoint, x::AbstractArray{C}, gamma=R(1)) where {R <: Real, C <: Union{R, Complex{R}}}
```
nantonel, line 49: `C<:RealOrComplex{R}`

src/functions/logBarrier.jl
```diff
@@ -29,55 +29,55 @@ end
 is_separable(f::LogBarrier) = true
 is_convex(f::LogBarrier) = true
 
-LogBarrier(a::T=1.0, b::T=0.0, mu::T=1.0) where {T <: Real} = LogBarrier{T}(a, b, mu)
+LogBarrier(a::R=1, b::S=0, mu::T=1) where {R, S, T} = LogBarrier{R, S, T}(a, b, mu)
 
 function (f::LogBarrier)(x::AbstractArray{T}) where T <: Real
     sumf = T(0)
     v = T(0)
```
nantonel, line 36: `zero(T)`

src/functions/logBarrier.jl
```diff
@@ -29,55 +29,55 @@ end
...
+    return -f.mu * sumf
 end
 
-function prox!(y::AbstractArray{T}, f::LogBarrier, x::AbstractArray{T}, gamma::Real=1.0) where T <: Real
-    par = 4*gamma*f.mu*f.a*f.a
+function prox!(y::AbstractArray{T}, f::LogBarrier, x::AbstractArray{T}, gamma::Real=1) where T <: Real
+    par = 4 * gamma * f.mu * f.a * f.a
     sumf = T(0)
     z = T(0)
     v = T(0)
```
nantonel, line 51: `zero(T)`

src/functions/logBarrier.jl
```diff
@@ -29,55 +29,55 @@ end
...
-    return -f.mu*sumf
+    return -f.mu * sumf
 end
 
 function prox!(y::AbstractArray{T}, f::LogBarrier, x::AbstractArray{T}, gamma::AbstractArray) where T <: Real
-    par = 4*f.mu*f.a*f.a
+    par = 4 * f.mu * f.a * f.a
     sumf = T(0)
     z = T(0)
     v = T(0)
```
nantonel, line 65: zero(T)

src/functions/logBarrier.jl
```diff
@@ -29,55 +29,55 @@ end
...
+        v = (z + sqrt(z * z + par_i)) / 2
+        y[i] = (v - f.b) / f.a
         sumf += log(v)
     end
-    return -f.mu*sumf
+    return -f.mu * sumf
 end
 
 function gradient!(y::AbstractArray{T}, f::LogBarrier, x::AbstractArray{T}) where T <: Real
     sum = T(0)
```
nantonel, line 77: zero(T)

src/functions/logisticLoss.jl
```diff
@@ -33,11 +33,11 @@ is_prox_accurate(f::LogisticLoss) = false
 
 # f(x)  =  mu log(1 + exp(-y x))
 
-function (f::LogisticLoss{T, R})(x::AbstractArray{R}) where {T, R}
+function (f::LogisticLoss)(x::AbstractArray{R}) where R
     val = R(0)
```
nantonel, line 37: zero(R)

src/functions/logisticLoss.jl
```diff
@@ -47,12 +47,12 @@ end
 #
 # Lipschitz constant of gradient: (mu y)
 
-function gradient!(g::AbstractArray{R}, f::LogisticLoss{T, R}, x::AbstractArray{R}) where {T, R}
+function gradient!(g::AbstractArray{R}, f::LogisticLoss, x::AbstractArray{R}) where R
     val = R(0)
```
nantonel, line 51: zero(R)

src/functions/normL0.jl
```diff
@@ -24,30 +24,30 @@ struct NormL0{R <: Real} <: ProximableFunction
...
-    return f.lambda*count(v -> v != 0, x)
+function (f::NormL0)(x::AbstractArray{T}) where {R, T <: RealOrComplex{R}}
+    return f.lambda * R(count(v -> v != 0, x))
 end
 
-function prox!(y::AbstractArray{T}, f::NormL0, x::AbstractArray{T}, gamma::Real=1.0) where T <: RealOrComplex
-    countnzy = 0
-    gl = gamma*f.lambda
+function prox!(y::AbstractArray{T}, f::NormL0, x::AbstractArray{T}, gamma::Real=1) where {R, T <: RealOrComplex{R}}
+    countnzy = R(0)
```
nantonel, line 34: `zero(R)`

src/functions/normL21.jl
```diff
@@ -28,73 +28,73 @@ end
...
+            nslice = R(0)
+            for j = 1:size(X, 2)
+                nslice += abs(X[i, j])^2
             end
             nslice = sqrt(nslice)
             scal = 1-gl/nslice
-            scal = scal <= 0.0 ? 0.0 : scal
-            for j = 1:size(X,2)
-                Y[i,j] = scal*X[i,j]
+            scal = scal <= 0 ? R(0) : scal
```
nantonel, line 82: zero(R)

src/utilities/normdiff.jl
```diff
@@ -1,8 +1,10 @@
 """
 Fast (non-allocating) version of norm(x-y,2)^2
 """
-function normdiff2(x,y)
-    s = 0.0
+function normdiff2(x::AbstractArray{C}, y::AbstractArray{C}) where {
+    R <: Real, C <: Union{R, Complex{R}}
+}
+    s = R(0)
```
nantonel, line 7: zero(R)

nantonel (approved):
All minor comments (just style). Feel free to ignore them. Great work! 😄

src/functions/logisticLoss.jl
lostella: I actually prefer `R(0)` to `zero(R)`, because the former can be used for any number

nantonel: yep, I imagine they're equivalent...

src/functions/huberLoss.jl
lostella: Ah! So I’ve actually changed my mind about that: `RealOrComplex{R}` is 17 characters while `Union{R, Complex{R}}` is 20, not much of a shorthand plus the latter is straight from `Base` and fully explicit

PR #94, opened 2020-03-06
Make prox on QuadraticDirect type stable

baggepinnen:
`Cholesky{R}` is an abstract type if the second type parameter is not specified. With this fix, a significant speedup is obtained for small Q where the dynamic dispatch is relatively expensive.

### DIFF
--- a/src/functions/quadraticDirect.jl
+++ b/src/functions/quadraticDirect.jl
@@ -25,7 +25,7 @@ function QuadraticDirect(Q::M, q::V) where {R <: Real, I <: Integer, M <: Sparse
 end
 
 function QuadraticDirect(Q::M, q::V) where {R <: Real, M <: DenseMatrix{R}, V <: AbstractVector{R}}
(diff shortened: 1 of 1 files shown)


src/functions/quadraticDirect.jl
```diff
@@ -25,7 +25,7 @@ function QuadraticDirect(Q::M, q::V) where {R <: Real, I <: Integer, M <: Sparse
 end
 
 function QuadraticDirect(Q::M, q::V) where {R <: Real, M <: DenseMatrix{R}, V <: AbstractVector{R}}
-    QuadraticDirect{R, M, V, Cholesky{R}}(Q, q)
+    QuadraticDirect{R, M, V, Cholesky{R, typeof(Q)}}(Q, q)
```
nantonel: perhaps `QuadraticDirect{R, M, V, Cholesky{R, M}}(Q, q)` would fit better?

baggepinnen: Absolutely!

PR #102, opened 2020-04-11
Optimizing projection on positive semidefinite cone

lostella:
Given the cost of the eigenvalue decomposition, this is expected to make a difference for small n. I'm coming up with some benchmark (perhaps I'll push them to ~~a dedicated package~~ the `benchmark` directory, and we could start running PkgBenchmark in CI), then I will add some results here.


lostella:
Benchmarks confirm slight improvement (up to 20% in some case) and no regression. See logs from CI/Benchmark workflow for now, we’ll find a way to make the benchmark results more easily accessible.

PR #108, opened 2020-05-01
Add indicator of Stiefel manifold, fix approximate (in)equalities, fix is_support

### LINKED ISSUE #107: Projection on Stiefel manifold
Have you guys ever considered implementing the Euclidean projection on the Stiefel manifold? See for instance, Section IV in https://people.eng.unimelb.edu.au/jmanton/pdf/Manton_Opt_Alg.pdf

Orthogonality constraints are very useful in low rank approximation and many many other applications. It would be nice to see how your (our) algorithms compare, for instance, with manifold optimization methods which are typically used to solve such problems.

lostella:
Fixes #107 by adding the indicator of the Stiefel manifold.

Fixes is_support to be based on the new is_positively_homogeneous (this is more accurate: for instance, the negative L1 norm is positively homogeneous, but is not the support function of a convex set; I’m not sure about *nonconvex* sets though).

Adds an approximate inequality utility, and tries to fix approximate equalities and inequalities in the evaluation of various functions, in an attempt to get rid of the annoying failing tests.


src/functions/indStiefel.jl
```diff
@@ -0,0 +1,42 @@
...
+    U_sliced = view(F.U, :, 1:p)
+    mul!(Y, U_sliced, F.Vt)
+    return R(0)
+end
+
+function prox_naive(f::IndStiefel, X::AbstractMatrix{T}, gamma::R=R(1)) where {R, T <: Union{R, Complex{R}}}
+    n, p = size(X)
+    F = svd(X)
+    Y = F.U[:, 1:p] * F.Vt
+    return Y, R(0)
```
lostella, line 41: @panpat let me know if this implementation looks reasonable

panpat: Looks fine! I wonder what happens if X is **not** of rank p...

lostella: If that's because of p > n, then an error will be raised (U cannot be sliced). Otherwise I'm not sure.

Maybe related: I guess the projection is unique if all singular values are distinct, otherwise different permutations of the singular spaces should result different projections (all at the same distance from x).

PR #118, opened 2020-11-28
Update and rename rpca.jl to rpcp.jl

caseykneale:
Isn't this principal component pursuit and not PCA? Awesome example

### DIFF
--- a/demos/rpcp.jl
+++ b/demos/rpcp.jl
@@ -1,4 +1,4 @@
-# rpca.jl - Robust PCA solvers using ProximalOperators
+# rpca.jl - Robust Principal Component Pursuit solvers using ProximalOperators
 #
 #   minimize 0.5*||A - S - L||^2_F + lam1*||S||_1 + lam2*||L||_*
 #


lostella:
Hi! Do you have a reference for this? If I'm not wrong, this example comes from Sec. 7.2 of https://stanford.edu/~boyd/papers/pdf/prox_algs.pdf but I'm happy to rename this in case "pursuit" is a better name and "analysis" refers to something else.

Also changing the filename here requires updating tests: https://github.com/kul-forbes/ProximalOperators.jl/blob/29d2e3a1369775e94fb49c2e223e771d80852a16/test/runtests.jl#L140

caseykneale:
https://www.researchgate.net/publication/258885717_Robust_PCA_via_Principal_Component_Pursuit_A_Review_for_a_Comparative_Evaluation_in_Video_Surveillance

So it seems like they are solving PCA via PCP so the original is fine. Sorry! I was under the impression this was just called PCP and not PCA, but apparently my memory isn't so good.

PR #122, opened 2021-01-17
Fix unbound args for Julia 1.6

lostella:
This fixes some issues with type parameters, that fail a test in Julia 1.6.0-beta1.

### DIFF
--- a/src/functions/indBinary.jl
+++ b/src/functions/indBinary.jl
@@ -13,7 +13,7 @@ S = \\{ x : x_i = low_i\\ \\text{or}\\ x_i = up_i \\},
 ```
 Parameters `low` and `up` can be either scalars or arrays of the same dimension as the space.
 """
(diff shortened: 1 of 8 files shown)


nantonel:
Looks good to me, simplifying subtyping is always a good idea 😄

src/functions/indBinary.jl
```diff
@@ -13,7 +13,7 @@ S = \\{ x : x_i = low_i\\ \\text{or}\\ x_i = up_i \\},
 ```
 Parameters `low` and `up` can be either scalars or arrays of the same dimension as the space.
 """
-struct IndBinary{C <: RealOrComplex, T <: Union{C, AbstractArray{C}}, S <: Union{C, AbstractArray{C}}} <: ProximableFunction
```
nantonel, line 16: wow this was strict!

nantonel (approved):

src/functions/indBinary.jl
lostella: yes :-) there's a lot of overly strict constraints, I fixed a few for now that are necessary to have tests pass on 1.6, we'll fix more later

PR #128, opened 2021-08-03
Add 1D-Total variation prox

### LINKED ISSUE #124: TV norm prox
Have you included the prox of the total variation norm, i.e. sum_i |x_i - x_{i-1}|?

I have implemented Condats algorithm in order to compute this (https://lcondat.github.io/publis/Condat-fast_TV-SPL-2013.pdf).
I could share my code.

Cheers!

fabian-sp:
I created the main file. Can you help me with what else I need to add? Also I am not familiar to programming in Julia, so if there are any best practices I am violating, let me know.


src/functions/normTV.jl
```diff
@@ -0,0 +1,111 @@
+# Total variation norm (times a constant)
+
+export NormTV
+
+"""
+**``TV`` norm**
```
lostella: Maybe you could mention “1-dimensional” here so that it’s immediately clear.

src/functions/normTV.jl
```diff
@@ -0,0 +1,111 @@
...
+
+NormTV(lambda::R=1) where {R <: Real} = NormTV{R}(lambda)
+
+function (f::NormTV)(x::AbstractArray)
+    return f.lambda * norm(x[2:end] - x[1:end-1], 1)
+end
+
+# Condat algorithm
+# https://lcondat.github.io/publis/Condat-fast_TV-SPL-2013.pdf
+function condat(y::AbstractArray, x::AbstractArray, lambda::Real)
```
lostella: `condat` here looks like a strange name to me: maybe `tvnorm_prox_condat` is more informative?

src/functions/normTV.jl
```diff
@@ -0,0 +1,111 @@
...
+                umax = -lambda; kplus = k;
+            end
+        end
+    end
+end
+
+function prox!(y::AbstractArray{T}, f::NormTV, x::AbstractArray{T}, gamma::Real=1.0) where T <: Real
+    a = gamma * f.lambda
+    y = condat(y, x, a)
+    return y
```
lostella: This function should return the value of `f` at `y` as output. In many cases, this is cheaply obtained as a by-product of the prox computation itself, but I’m not sure this is the case here: worst case, you can just compute it from scratch.

lostella (changes requested):
Nice! Thanks for this addition! I left some comments inline.

I think two things are missing:
- An entry for this feature in the documentation (just add the type name eg below this line https://github.com/JuliaFirstOrder/ProximalOperators.jl/master/docs/src/functions.md#L67)
- A test, for which you could eg add one or more test cases to this test file https://github.com/JuliaFirstOrder/ProximalOperators.jl/blob/master/test/test_optimality_conditions.jl This requires implementing a check for the optimality conditions of the prox problem: these are in eqn (7) in the paper, I think they can be evaluated starting from x* (the output point in the code) and using eqns (5) and (6) too. Unless I’m missing something?

src/functions/normTV.jl
lostella: Errata: `f(y)` should be the *only* output (edited previous post)

src/functions/normTV.jl
fabian-sp: done

src/functions/normTV.jl
fabian-sp: done

src/functions/normTV.jl
fabian-sp: done

fabian-sp:
I added now a test for the optimality conditions. Worked for me for a couple of test cases, though the approximately equal to 0 assertion is a bit tricky. If you have any suggestions let me know.

src/functions/normTV.jl
```diff
@@ -0,0 +1,112 @@
...
+** 1-dimensional ``TV`` norm**
+
+    NormTV(λ=1)
+
+With a nonnegative scalar parameter λ, returns the function
+```math
+f(x) = λ ∑_{i=2}^{n} |x_i - x_{i-1}|.
+```
+"""
+struct NormTV{T <: Real} <: ProximableFunction
```
lostella: On a second thought: I’m not sure having “norm” in the type name is ideal, since this is technically not a norm: for example, `f(x) == 0` does not imply `x .== 0`. Maybe a better name for this type is just `TotalVariation` or `TotalVariation1D`, what do you think? If so, then the file name and other references to this in the code should change accordingly

test/test_optimality_conditions.jl
```diff
@@ -21,6 +21,25 @@ check_optimality(f::IndBallL1, x, gamma, y) = begin
...
+    c1 = all(abs.(u) .<= gamma*f.lambda + 1e-10)
+    # check whether last equals 0 (first by construction)
+    c2 = isapprox(u[end], 0, atol=1e-12)
+    # check whether equal +- gamma*lambda
+    h = sign.(y[1:end-1] - y[2:end])
+    c3 = all(isapprox.( u[2:end-1] .* abs.(h) , h *f.lambda*gamma))
+    return c1 && c2 && c3
+end
+
 test_cases = [
```
lostella, line 44: You also need to add some test cases to this list (analogously to the other ones here, just using the total variation function), so that we can look deeper into the failing checks

test/test_optimality_conditions.jl
```diff
@@ -21,6 +21,25 @@ check_optimality(f::IndBallL1, x, gamma, y) = begin
     return all(sign_is_correct) && check_optimality(IndSimplex(f.r), abs.(x), gamma, abs.(y))
 end
 
+check_optimality(f::NormTV, x, gamma, y) = begin
+    N = length(x)
+    # compute dual solution
+    u = zeros(N+1)
```
lostella: You could enforce `u` this to have elements of the same type as `x` (or `y`). To do this you can do

```julia
R = real(eltype(x))  # or y
u = zeros(R, N+1)
```

test/test_optimality_conditions.jl
```diff
@@ -21,6 +21,25 @@ check_optimality(f::IndBallL1, x, gamma, y) = begin
...
+    u[1] = 0
+    for k in 2:N+1
+        u[k] = x[k-1]-y[k-1]+u[k-1]
+    end
+
+    # check whether all duals in interval
+    c1 = all(abs.(u) .<= gamma*f.lambda + 1e-10)
+    # check whether last equals 0 (first by construction)
+    c2 = isapprox(u[end], 0, atol=1e-12)
```
lostella: Tolerances here could be dependent on the float type (`R` following the comment above). For example, they could be multiples of `eps(R)` or depend on it in some other way.

lostella (changes requested):
I almost forgot: to be able to use this, you need to `include` the source file from the main package script: https://github.com/JuliaFirstOrder/ProximalOperators.jl/blob/master/src/ProximalOperators.jl this is why tests are currently failing (the new type cannot be used)

How big of an issue is the assertion on `u[end]`? Does it fall significantly away from zero? Maybe you could add some (non random) test cases where the problem occurs, to better look into it

src/functions/normTV.jl
fabian-sp: Yes, I think you are right. I change it as proposed.

test/test_optimality_conditions.jl
fabian-sp: I used `10*eps(R)` now.

src/functions/TotalVariation1D.jl
```diff
@@ -0,0 +1,112 @@
...
+
+function (f::TotalVariation1D)(x::AbstractArray)
+    return f.lambda * norm(x[2:end] - x[1:end-1], 1)
+end
+
+# Condat algorithm
+# https://lcondat.github.io/publis/Condat-fast_TV-SPL-2013.pdf
+function tvnorm_prox_condat(y::AbstractArray, x::AbstractArray, lambda::Real)
+    # solves y = arg min_z lambda*sum_k |z_{k+1}-z_k| + 1/2 * ||z-x||^2
+    N = length(x);
```
lostella: The semicolons are not needed, they are usually only used to put multiple expressions in a single line I think

lostella (commented):
First of all, good job and thanks for translating the algorithm from the paper into a GOTO-free version, that must have been a bit of a puzzle!

It looks like the test cases are not covering many of the algorithm branches (at least according to codecov’s comments). This may be due to the fact that the test cases are very small or not particularly relevant for this algorithm (like in case `y` ends up being equal to `x`, or a constant vector). One way to come up with better ones might be to:
- design a piecewise-constant signal, with a handful of change points, of large enough magnitude so that denoising doesn’t get rid of them
- add small magnitude noise, that would be killed by denoising

This way one should be able to have fixed (non-random) test instances of small size. 
To nail what “small vs large” magnitudes are, one can do trial and error or try to understand this from the optimality conditions or algorithm logic, whatever is easier.

Sorry for this further ask :-) but I think it’s worth having test cases that cover the algorithm properly. If I have time I’ll try to come up with some test cases too!

fabian-sp:
So I played around a bit and it seems that some of the `if`-clauses (the first two) are only for edge cases (which I havent found yet).
Something like
`x = vcat(LinRange(1., -1., 10), -1*ones(3), LinRange(-1., 1., 10))` works okay as it runs through a couple of different clauses. But we will not get 100% coverage with that. I'll try a bit more.

fabian-sp:
@lostella I added the "pathological case" from section 2A in the paper. Together with the other test cases, I think almost every clause but the first is now covered. But I don't really know how to get the last one. Did you have any success?

lostella:
> Did you have any success?

Thanks! Didn’t get to it yet, but maybe it’s enough test cases already. I’ll let the tests run and if everything is green approve & merge

lostella (approved):

PR #133, opened 2021-09-06
Add NormL1plusL2

fabian-sp:
I created the function for the sum of L1 and L2. I think together with `SlicedSeparableSum` this is exactly what you need for the Group Lasso penalty.
Regarding tests: could add that it has to match `NormL1` and `NormL2` for one of the two parameters being zero, but was not so sure how/where to do that.
Another question: I had some problems importing the `prox!`-function from `NormL1/2`, and solved it with 

`import ProximalOperators: prox!`

Is this the intended way in Julia to import functions from other files within a module?

Cheers,
Fabian


docs/src/functions.md
```diff
@@ -66,6 +66,7 @@ NormLinf
 NuclearNorm
 SqrNormL2
 TotalVariation1D
+L1plusL2
```
lostella: I think it’s nice to keep these in alphabetical order, the others should be at least

src/functions/L1plusL2.jl
```diff
@@ -0,0 +1,56 @@
...
+function (f::L1plusL2)(x::AbstractArray)
+    return f.lambda1 * norm(x, 1) + f.lambda2 * norm(x, 2)
+end
+
+function prox!(y::AbstractArray{T}, f::L1plusL2, x::AbstractArray{T}, gamma::Real=1) where T <: Real
+
+    f1 = NormL1(f.lambda1)
+    f2 = NormL2(f.lambda2)
+
+    y0, t = prox_naive(f1, x, gamma)
```
mfalt: `prox_naive` is generally intended to be a simple implementation used to test the functionality, but it is also often inefficient, so we don't want to use it in our `prox!` implementations.

The simple way to solve this is to do
```julia
y0 = similar(y)
prox!(y0, f1, x, gamma)
```

However, the implementation on prox! on NormL2 allows aliasing, so even better is simply
```julia
prox!(y, f1, x, gamma)
prox!(y, f2, y, gamma)
```

src/functions/L1plusL2.jl
```diff
@@ -0,0 +1,56 @@
+# (weighted) sum of L2 norm and L1 norm
+# for Group Lasso, this can be used together with src/calculus/slicedSeparableSum
+
+using ProximalOperators
+import ProximalOperators: prox!
```
mfalt: using and import would only be needed of you try to run this file separately from the package, which might have been the case when you tested it. But it is not needed since the code already "lives" in the module now.

mfalt (changes requested):

src/functions/L1plusL2.jl
fabian-sp: Thanks! I was not sure whether having `y` as input and writing the result  into `y` could potentially cause problems. But then I will change it.

src/functions/L1plusL2.jl
fabian-sp: I am not so experienced with writing Julia packages. So I can always use any functions from inside the module without any import statement (except maybe for the testing)?

src/functions/L1plusL2.jl
mfalt: You are right that it can be dangerous in general, but in this case it will only be called on NormL2, and the implementation here looks safe in that regard:
https://github.com/JuliaFirstOrder/ProximalOperators.jl/blob/c9a250f8836e41450bada594ed23baa22dc8e928/src/functions/normL2.jl#L39

src/functions/L1plusL2.jl
mfalt: Yes, when you are inside the module (which you remain in when including the file in the module), the functions are all in that "namespace" already. The tests are not in the same module, so there it is needed to import.

src/functions/L1plusL2.jl
```diff
@@ -0,0 +1,52 @@
+# (weighted) sum of L2 norm and L1 norm
+# for Group Lasso, this can be used together with src/calculus/slicedSeparableSum
+export L1plusL2
+
+"""
+**L1 norm plus L2 norm**
+
+    L1plusL2(λ_1=1, λ_2=1)
```
lostella: This is technically a norm, right? Since it's a linear combination of norms. Then you could name it `NormL1plusL2` for consistence with other norms

lostella (changes requested):

mfalt:
I wasn't completely happy with the implementation of `prox!` so I made a suggestion in a PR to your branch: https://github.com/fabian-sp/ProximalOperators.jl/pull/1

> I didn't like that we created the L1 and L2 norms in the `prox!` call because it forces us to recheck that the parameters are positive every time. I rewrote the struct to contain the norm functions instead, and allowed `lambda1` to be a vector. I think the code looks slightly better this way.
> This PR also defer almost all computations to existing implementations.

mfalt:
I made a couple of changes, for example allowing complex domain, added more tests, and fixed the printing for Array case. I hope you don't mind.
I am not too happy about my implementation of `prox_naive` do we have a better way to verify all cases?
Can someone verify that the same rule works for complex arguments and vector `lambla1`?

fabian-sp:
Hi Mattias, I am not sure whether the proof would work for complex `lambda1` as it involves the subdifferential, and I have never encountered that for complex-valued function.

nantonel:
Not completely sure, but perhaps prox_naive could be formed using calc rule regularise applied to norm1?

fabian-sp:
Hi all, is this still open/ under discussion?

Best, Fabian

lostella:
@fabian-sp this looks good to me! Thanks for implementing this!

lostella (approved):

lostella:
> Hi Mattias, I am not sure whether the proof would work for complex `lambda1` as it involves the subdifferential, and I have never encountered that for complex-valued function.

I think the general setting of the package is that of (extended) real valued functions, so all coefficients in front of norms and penalties should have no imaginary part

PR #135, opened 2021-12-02
Fixes to IndHyperslab, bump Julia version to 1 and 1.6 in workflows

lostella:


test/test_indHyperslab.jl
```diff
@@ -1,13 +1,10 @@
...
 using ProximalOperators
-using Random
 using Test
 
-Random.seed!(0)
-
 @testset "IndHyperslab" begin
 
-for R in [Float16, Float32, Float64]
+for R in [Float32, Float64]
```
lostella, line 7: @nantonel I just removed Float16 here since one check we do is a bit tricky for indicator functions, especially with low precision

lostella (commented):
I’ll go ahead and merge this since it’s an overall minor change

PR #150, opened 2025-02-21
Added Unit Simplex

matbesancon:
Couldn't find a better name, that's the one we use in FrankWolfe.jl.

Having a dedicated prox avoids having to create a n+1-dimensional simplex

### DIFF
--- a/src/functions/indSimplex.jl
+++ b/src/functions/indSimplex.jl
@@ -1,6 +1,6 @@
 # indicator of a simplex
 
-export IndSimplex
+export IndSimplex, IndUnitSimplex
 
 """
     IndSimplex(a=1.0)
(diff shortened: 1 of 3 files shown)


src/functions/indSimplex.jl
```diff
@@ -107,3 +107,49 @@ function prox_naive(f::IndSimplex, x, gamma)
     end
     return v, R(0)
 end
+
+"""
+    IndUnitSimplex(a=1.0)
+
+Return the indicator of the unit simplex
+```math
+S = \\left\\{ x : x \\geq 0, \\sum_i x_i \\leq a \\right\\}.
```
lostella, line 116: Hmmm yeah, naming is tough here. Would this set be the convex hull of S u {0}, where S is that of IndSimplex? (i.e. with equality on the sum)

matbesancon:
Yes exactly that's one way to define it

src/functions/indSimplex.jl
lostella: Well, the origin 0 \in R^k is the projection to R^k of e_{k+1} \in R^{k+1}. So I guess this set is the projection to R^k of the k-simplex (which lives in R^{k+1} originally). If so, this could suggest a better name maybe.

Am I making sense? Works in my mind for k=2 at least (hehe).

PR #151, opened 2025-11-11
Move traits to ProximalCore, new functions, RecursiveArrayToolsExt, defer loading OSQP

hakkelt:
This PR is part of the effort to make StructuredOptimization as general as possible (see the PR on OperatorCore: https://github.com/JuliaFirstOrder/ProximalCore.jl/pull/5).

Changes:
- `PrecomposedSlicedSeparableSum`: function to be used for complex problems with multiple variables and variables sliced in different terms. E.g.:
```julia
x, y = Variable(10), Variable(5)
problem = ls(x[1:5] - y) + norm(x, 1) + norm(x[6:10], 1)
```
- `reshapeInput` wrapper can help, for example, in low-rank minimization to avoid reshaped `Eye` from AbstractOperators and its unnecessary copy:
```julia
x = Variable(3,3,5)
problem = norm(reshape(x, 9, 5), *)
```
- `RecursiveArrayToolsExt` extension contains some functions that were moved from StructuredOptimization as I believe they belong here.
- Loading OSQP dependency was delayed until `IndPolyhedralOSQP` is instantiated. The reason behind it was that OSQP is one of the largest dependencies of ProximalOperators, and it is only used by a single function. ProximalOperators loads almost a second faster without it.


hakkelt:
I'll break this large PR into multiple smaller ones, but I'll keep this open for reference.

hakkelt:
This PR is no longer needed because all of its changes are either merged in separate PRs or have a currently open PR to implement them (#162).

PR #156, opened 2026-02-02
Add reshapeInput function wrapper

hakkelt:
This calculus-type function is useful when the norm is already implemented, but it requires a specifically shaped input. For example, I work with 4D image arrays (3D medical images that change over time in the fourth dimension), and if I want to apply the proximal operator of the nuclear norm, I need to reshape them into a matrix before the operation.


hakkelt:
Hi @lostella, could you look at this PR?

lostella:
@hakkelt thanks for the reminder! The wrapper looks quite generally useful and simple. Tests look good!

lostella (approved):

lostella:
If there are no additional changes you would like to add, I would merge this

hakkelt:
Thanks, I don't want to add anything else.

PR #157, opened 2026-04-13
Add PrecomposedSlicedSeparableSum implementation and corresponding tests

hakkelt:
This new function implements the intersection/composition of Precompose and SlicedSeparableSum. The long name suggests a code smell, but I could not find a cleaner way to implement this composition, so I'm open to better solutions. The goal is to implement prox for tricky expressions like the following:
```julia
x = (randn(10), randn(10))
norm(x[1], 1) + norm(A2[1:5, 1:5] * x[2][1:5], 2) + norm(A2[6:10, 6:10] * x[2][6:10], 2)^2
```


hakkelt:
@lostella: Could you review this PR?

The benchmarking is still failing because of the lack of permission. :disappointed:  My guess is that adding write permisison to workflows (Settings -> Actions -> General -> Workflow permissions -> Read repository contents and packages permissions) might solve the problem. I'm not really expert on security, so I'm not sure how dangerous it is exactly. What I read about that is that it can lead to exposure of secrects (e.g. if someone creates a PR that modifies the benchmarking script such that it sends the GITHUB_TOKEN to an external server). But as your approval is required for PRs of new contributors by the current settings, the threat is greatly reduced, I think.

The other option could be to disable commenting and push the results only to job summary (https://astroautomata.com/AirspeedVelocity.jl/stable/#Option-2:-Job-Summary).

lostella:
@hakkelt will review asap.

I agree that probably disabling comments from the benchmarking workflow, and logging the result instead, is the safest at the moment. Does the workflow fail in case of performance regressions? If so, that should be enough: red => inspect logs

hakkelt:
@lostella I'll fix the benchmarking problem in a separate PR.

hakkelt:
Is there anything I can do for this PR or help the reviewing process?
Btw, the benchmarking results are available here, and they seem to be ok (which is expected as this PR only adds features but doesn't modify any existing functions): https://github.com/JuliaFirstOrder/ProximalOperators.jl/actions/runs/24719301336

lostella (approved):
@hakkelt looks good! Sorry, fell off my radar!

hakkelt:
Thanks!

PR #158, opened 2026-04-13
Benchmarking results as job summary

hakkelt:
Let's see what it looks like if benchmarking results are saved as a job summary...

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,8 +1,7 @@
 name: Benchmark
 on:
   - pull_request
-permissions:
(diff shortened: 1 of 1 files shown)


lostella:
I think it looks ok. Needs clicking around to discover it, but it’s fine. My guess though is that the workflow will never fail, even in case of regressions?

hakkelt:
Yes, the job only fails in case of error. I'm thinking about appending a small script to the workflow that checks the results (possibly I'll refurbish some parts from the previous benchmarking solution), and throws error when regressions are detected.

lostella (approved):

PR #159, opened 2026-04-15
fix benchmark result posting

hakkelt:
In this PR, I try to fix this problem by dividing the benchmarking into two phases:
1. Run the benchmark, save the results to a file, and upload it as an artifact. -> It does not need any special permissions.
2. Download the artifact and post the content as a comment. -> This gets triggered by the first phase via "workflow_run", and my hope is that it is possible to selectively set write permission to this job only.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,13 +1,17 @@
 name: Benchmark
 on:
   - pull_request
-permissions:
-  pull-requests: write    # needed to post comments
+
+# No special permissions needed — the action uploads an artifact
+# instead of commenting directly. A separate workflow_run workflow
+# (post_benchmark_comment.yml) picks up the artifact and posts the comment.
+
 jobs:
   benchmark:
     name: Benchmark
     runs-on: ubuntu-latest
     steps:
-      - uses: MilesCranmer/AirspeedVelocity.jl@action-v1
+      - uses: hakkelt/AirspeedVelocity.jl@master
         with:
           julia-version: '1'
+          post-comment: 'false'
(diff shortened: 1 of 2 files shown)


hakkelt:
To test this PR, post_benchmark_comment.yml needs to be on master already... So I'll play around a bit in some playground repositories, then I'll report if it works.

hakkelt:
I opened a PR on AirspeedVelocity (https://github.com/MilesCranmer/AirspeedVelocity.jl/pull/142) that adds this two-stage benchmarking workflow.

@lostella  For now, there are three options, I think:
- Merge the other benchmarking PR that saves the results into the job summary (https://github.com/JuliaFirstOrder/ProximalOperators.jl/pull/158)
- Merge this PR that uses my fork and posts comments safely
- Wait until my PR on AirspeedVelocity gets merged, then update this PR to use the new version

Regarding the automatic detection of performance regressions, I considered it and concluded that making the job fail might not be the best solution. I opened another PR on AirspeedVelocity (https://github.com/MilesCranmer/AirspeedVelocity.jl/pull/140) that adds emojis to its result table to mark significant speedups and slowdowns (actually, I just re-heated the original author's outdated PR). If it gets merged, I'm considering opening another one that adds a notice before the collapsible table in the posted comment, warning about the negative performance regressions.

lostella:
@hakkelt thank you. I would probably merge #158, and then come back to this if support from AirspeedVelocity is added. What do you think?

hakkelt:
I agree. That would be a clean approach that still allows the detection of performance regressions.

hakkelt:
This PR turned out be unnecessary as https://github.com/JuliaFirstOrder/ProximalOperators.jl/pull/161 fixed the problem more cleanly.

Issue #163, opened 2026-08-19
Heavy dependency: OSQP


hakkelt:
# Problem

The loading time is ProximalOperators is surprisingly high (at least for me with 2.77 s). Examining the problem I realized that OSQP is a dependency used by a single proximal operator (IndPolyhedralOSQP), but loading it (and its dependencies) accounts for ~2s out of the ~2.77s it takes to load ProximalOperators. Anyone who doesn't need IndPolyhedralOSQP still pays this cost on every using ProximalOperators.

I'd like to fix this, and see three possible approaches. I'm happy to implement whichever one @lostella prefers — dropping #162 is completely fine if a different approach is favored.

## Option 1 — Defer loading at runtime (implemented in #162)

Keep the current API (IndPolyhedralOSQP always available, OSQP stays a hard dependency), but only actually load/require the OSQP module the first time it's needed, via Base.require + Base.invokelatest:
```julia
function IndPolyhedralOSQP(l::AbstractVector{R}, A::AbstractMatrix{R}, u::AbstractVector{R}) where R
    mod = Base.invokelatest(Base.require(@__MODULE__, :OSQP)) do OSQP
        mod = OSQP.Model()
        OSQP.setup!(mod; P=SparseMatrixCSC{R}(I, n, n), l=l, A=sparse(A), u=u, ...)
        mod
    end
    new{R,typeof(mod)}(l, A, u, mod)
end
```
**Pros**: zero API breakage, OSQP still "just works" without any extra using.
**Cons**: this is essentially working around Julia's package-loading model rather than using it — it forces runtime dispatch (invokelatest) on every call that touches OSQP, which adds overhead and defeats specialization/inlining. It's a bit of an antipattern; I only reach for it because it doesn't change what users have to do.

## Option 2 — Package extension (weak dependency)

Move OSQP to a weak dependency and implement IndPolyhedralOSQP in a package extension (ext/ProximalOperatorsOSQPExt.jl), the way SparseArrays/ChainRulesCore-style extensions work in general. Without OSQP loaded, calling IndPolyhedralOSQP throws an informative error; loading OSQP (even just using OSQP alongside ProximalOperators) activates the extension and the constructor works normally.
```toml
# Project.toml
[weakdeps]
OSQP = "ab2f91bb-94b4-55e3-9ba0-7f65df51de79"

[extensions]
ProximalOperatorsOSQPExt = "OSQP"
```

```julia
[...]

PR #165, opened 2026-09-08
Move OSQP to a weak dependency (package extension)

hakkelt:
Implements "Option 2" from #163.

## Goal

OSQP and its dependencies add ~2s to `using ProximalOperators`, but OSQP is
only used by `IndPolyhedralOSQP` — the default backend of `IndPolyhedral`.
This makes OSQP optional so only users who need that operator pay the cost.

## Changes

- OSQP moved from `[deps]` to `[weakdeps]`, registered as the extension
  `ProximalOperatorsOSQPExt`.
- `src/functions/indPolyhedralOSQP.jl` is removed. The main package keeps only
  the `IndPolyhedralOSQP` type and a stub constructor that raises an
  informative error when OSQP is not loaded; both are now inlined into
  `src/functions/indPolyhedral.jl`.
- The extension holds everything that needs OSQP: the constructors, `prox!`,
  the function evaluation and `prox_naive` — i.e. every method operating on an
  `IndPolyhedralOSQP` instance.
- `IndPolyhedral` docstring documents the new requirement.

## Effect

Constructing `IndPolyhedral(...)` / `IndPolyhedral(...; solver=:osqp)` now
requires `using OSQP` first; without it, a clear error is thrown instead of
a `MethodError`. This is a minor breaking API change. All other operators are
unaffected and load without OSQP.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/functions/indPolyhedral.jl
```diff
@@ -23,6 +26,34 @@ function IndPolyhedral(args...; solver=:osqp)
...
+    u::AbstractVector{R}
+    mod::M
+    # Explicit inner constructor: suppresses the auto-generated 4-positional-arg
+    # outer constructor, which would otherwise shadow the `(l, A, xmin, xmax)`
+    # constructor added by ProximalOperatorsOSQPExt.
+    IndPolyhedralOSQP{R, M}(l, A, u, mod) where {R, M} = new{R, M}(l, A, u, mod)
+end
 
-include("indPolyhedralOSQP.jl")
+is_proximable(::Type{<:IndPolyhedralOSQP}) = false
```
lostella, line 50: Note: in light of the discussion in ProximalCore: this may need to become true.

But if other operators have this misuse, then a separate PR is fine too

lostella (approved):

src/functions/indPolyhedral.jl
hakkelt: I think it is better to handle this problem in a separate PR, so please merge this PR, and I'll open another for that.

